PR #76, opened 2022-10-13
Old-skool experience

### LINKED ISSUE #11: Make Predictive text mode smarter
**After #7**.
- ~~Changing the case is only possible while composing a word. "#" short press changes the case.~~ Done in #44 
- ~~Default text mode is lowercase.~~
- ~~Start with a capital letter (CAPS_SINGLE mode): ~~
  - ~~If the text field is empty~~
  - ~~At the beginning of a sentence (after `.`, `?`, `!`)~~
- ~~If possible, get rid of the extra icons for upper/lower/single case for predictive modes. Keep only one icon.~~ Done in #44

### LINKED ISSUE #51: DPAD_CENTER should be able to work as "submit" button
Try making the DPAD_CENTER button work as a submit or Enter button. [sendDefaultEditorAction()](https://developer.android.com/reference/android/inputmethodservice/InputMethodService.html#sendDefaultEditorAction(boolean)) could be useful. Alternatively, consider making use of [sendKeyChar()](https://developer.android.com/reference/android/inputmethodservice/InputMethodService.html#sendKeyChar(char)).

This seems to be important, because some forms do not have a "submit" button, so one should "press ENTER" to submit them.

### LINKED ISSUE #60: Accepting the suggestions in Predictive mode is inconsistent
Currently, there is no way of accepting a partially typed word. Also, Center key accepts suggestions inconsistently. So, let's follow the Nokia 3310 rules:

- ~~Center~~
  - ~~accepts the current database suggestion. Currently, one should press Left/Right, then Center. This is confusing and inconsistent.~~
  - ~~accepts the current generated suggestion (this works fine now).~~
- ~~Left: accepts the partially typed word as is and moves the cursor before it.~~
- Right: will be handled in #69 
- ~~Update the user readme.~~

### LINKED ISSUE #68: When there are no more suggestions in the database, suggest all possible "next" combinations instead of the first one only
For example, we are trying to type "touchscreen", but the suggestions in the database are up to "touch".

1. Type "86824" => "touch".
2. Now press "7", for "touchs".

This will result in: "touchp" as the only suggestion, because we currently select the first letter for 7. But it should instead propose all the possible combinations for 7: "touchp", "touchq", "touchr" and "touchs".

This will be of great benefit when trying to type unknown words, especially in combination with #69 .

### LINKED ISSUE #69: Allow selecting a suggestion stem in Predictive mode
At any point, while typing a word, the user may select a suggestion using the arrow keys. This will greatly benefit unknown word suggestion, especially in combination with #68.

- Once a suggestion is selected, set is as a stem. 
- Every subsequent keypress will result in suggestions only starting with the stem.
- The stem is cleared when:
  - regular `InputMode.reset()` is performed.
  -  accepting the current word.
  - `predictionSequence.length() < stem.length()`.

### LINKED ISSUE #77: No software SEND function - possible via hardware keys?
Hi, love the keyboard so far and want to start by saying that I am not an experienced Android dev.

I'm using the Xiaomi Qin F21 Pro and am using a few apps without a send button next to the text input. Using the enter key (when remapped) also does nothing. I'm guessing this is because the SEND function is a softkey function and can't be remapped to a hardware based key, as evidenced by people trying (and failing) to do so with Bluetooth keyboards on Android. Would it be possible to, near to or next to the predictive text at the bottom of the screen, add a software-based SEND function to work-around this issue? It's making certain text fields incredibly difficult to fill out - I type what I want in T9 then have to switch to the AOSP keyboard just to press send. 

Thanks.

sspanak:
Fixes #60 
Closes #68 
Closes #51 
Closes #77 
Closes #11 
Closes #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@ TT9 is an IME (Input Method Editor) for Android devices with hardware keypad. It
 
 This is an updated version of the [original project](https://github.com/Clam-/TraditionalT9) by Clam-.
 
+## Using Traditional T9
+If you just wish to install and use TT9, see the [user manual](docs/user-manual.md). You don't need to read anything below this line.
+
 ## Building
 The recommended way of building is using Android Studio. As the of time of writing this, the current version is: Android Studio Dolphin | 2021.3.1.
 
@@ -52,9 +55,6 @@ To support a new language one needs to:
 - Add a new entry in `res/values/const.xml`. Make sure the new ID matches the one in the language class.
 - Add new entries in `res/values/arrays.xml`.
 - Add translations in `res/values/strings-your-lang`. The Android Studio translation editor is very handy.
-
-## Using the app
-See the [user manual](docs/user-manual.md).
 
 ## Word Lists
 Here is detailed information and licenses about the word lists used:
--- a/docs/user-manual.md
+++ b/docs/user-manual.md
@@ -14,7 +14,7 @@ _The actual menu names may vary depending on your Android version and phone._
 ### Enabling Predictive Mode
 With the default settings, it is only possible to type in 123 and ABC modes. In order to enable the Predictive mode, there are additional steps:
 
-- Open the [Configuration screen](#configuration-options).
+- Open the [Settings screen](#settings-screen).
 - Select the desired languages.
 - Load the dictionaries.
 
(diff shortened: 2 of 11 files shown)


sspanak:
Readme section for UP and DOWN must be updated too

PR #84, opened 2022-10-18
Add Dutch language

TomJansen:


sspanak:
@TomJansen, there are a couple of things about the word list.

1. I see there are words such as:
```
06
06-nummer
06-nummers
```
However, in Predictive mode it is not possible to type words that contain numbers, because, well... numbers are not words. So please remove them from the file. They would only take up space, but would never appear as suggestions.

Besides, it is always possible to type any of the above by holding "0", then "6", then proceeding with the actual word. There is no need to have the numbers in the dictionary. 

2. Single letters are added automatically, so there is no need to include them either.

3. Are these real words?
```
z'en
z'je
z'n
z's
z'tje
z'tjes
z.g.
z.g.a.n.
z.i.
z.o.z.
z.s.m.
```
4. The ones below seem to be French or German words. Are they used in Dutch? If yes, is this the correct spelling? I am trying to understand if the extra characters are actually required. Having less, will make it easier both for the phone and for people, when typing. If the answers are "no", let's convert the words to Dutch spelling or remove them.
```
à
à-la-carterestaurant
à-la-carterestaurants
échéance
échéances
élégance
époque
één
öre
öres
über-ich
überhaupt
übermensch
übermenschen
```

TomJansen:
1. OK, I removed all words containing numbers
2. Removed those as well
3. In Dutch, those are abriviations, and used quite a lot!
4. These are loan words (however, not "één"), some of them are used more than others. However, extra characters are used in Dutch. Querying the official AOSP Dutch dictionary from https://android.googlesource.com/platform/packages/inputmethods/LatinIME/+/refs/heads/master/dictionaries/ (which has 3x less words btw) reveals this distribution of extra characters used: 
  É - 2
  Ö - 4
  á - 13
  â - 4
  ã - 1
  ä - 24
  ç - 21
  è - 172
  é - 329
  ê - 34
  ë - 1243
  í - 16
  î - 3
  ï - 483
  ñ - 4
  ó - 10
  ô - 4
  ö - 158
  ø - 2
  û - 3
  ü - 66

TomJansen:
I also used the wrong font for the taskbar icon, fixed that now

sspanak:
> 4\. These are loan words (however, not "één"), some of them are used more than others. However, extra characters are used in Dutch. Querying the official AOSP Dutch dictionary from https://android.googlesource.com/platform/packages/inputmethods/LatinIME/+/refs/heads/master/dictionaries/ (which has 2x less words btw) reveals this distribution of extra characters used:

I see. Well, my philosophy is: it is OK to have some loan words, but don't go over the top. Language X dictionary _is_ supposed to contain language X words. If you need to type in another language, just switch to that language.

Anyway, if you think you are all done, I'll double check everything, test on my phone just in case, then merge.

TomJansen:
Yes I think this PR is all done. We can always change the dictionary words later if it contains too many loan words, but I think it is OK. It is the same dictionary that is used by Firefox and OpenOffice for Dutch.

sspanak (approved):

PR #96, opened 2022-11-01
Rewrite candidate view

TomJansen:
This is my rewrite of the candidate view. However, I cannot seem to figure out how to highlight the candidate word after `scrollToPosition`. Also, this does not contain any onclicklisteners in the candidatebar, which could be implemented. But I am not sure how.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1,31 +1,31 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-	<string translatable="false" name="help_url">https://github.com/sspanak/tt9/blob/master/docs/user-manual.md</string>
-	<string name="app_name" translatable="false">Traditional T9</string>
-	<string name="app_settings">Traditional T9 Settings</string>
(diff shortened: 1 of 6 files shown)


sspanak:
@TomJansen, it seems your soultion is very much similar to mine. I've figured out how to highlight the current candidate, so I'll take care of that. Don't worry. Also, I would like to fine tune the spacing and the colors.

res/values/strings.xml
```diff
@@ -1,31 +1,31 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-	<string translatable="false" name="help_url">https://github.com/sspanak/tt9/blob/master/docs/user-manual.md</string>
```
sspanak, line 3: What happened here? I don't see any real changes or are there any? If not, let's revert the file and keep the diff minimal.

sspanak: You know, I'll revert it in my branch, while cleaning up the rest of the code.

sspanak (approved):

PR #120, opened 2022-11-27
Update README.md

Poussinou:
Hi,

This small PR adds a button to get your app via the [IzzyOnDroid open source app repo](https://apt.izzysoft.de/fdroid/) in a F-Droid client, and a link to get the APK from the Releases Section.

It will help people without Google Play to find and use your app, as well as helping them to keep it updated.

Since the apk provided by IzzyOnDroid is simply taken from the Release Section of this repo, please never forget to attach your signed apk to the release when tagging a new version, so that Izzy can take it and make it available via a F-Droid client to your users !

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,12 @@ TT9 is an IME (Input Method Editor) for Android devices with hardware keypad. It
 
 This is an updated version of the [original project](https://github.com/Clam-/TraditionalT9) by Clam-.
 
+[<img src="https://gitlab.com/IzzyOnDroid/repo/-/raw/master/assets/IzzyOnDroid.png"
+     alt="Get it on IzzyOnDroid"
+     height="80">](https://apt.izzysoft.de/fdroid/index/apk/io.github.sspanak.tt9)
+
+or get the APK from the [Releases Section](https://github.com/sspanak/tt9/releases/latest).
+
 ## Using Traditional T9
 If you just wish to install and use TT9, see the [user manual](docs/user-manual.md). You don't need to read anything below this line.
 


Poussinou:
@Izzysoft it's for you :)

IzzySoft:
> Does Izzy ingore the ones marked as "pre-release"?

Depends. Usually I ignore them, but that's an option that can be set per-app. Let me check how it's set for yours… Yes, pre-releases are skipped. And I see you're now using sane tag and version names, so I can match tags based on them? My updater uses [semantic versioning](https://semver.org/) for that based on `versionName` or numeric comparison based on `versionCode`, depending on what tag names are used. In your case, I had to disable that and hope for the date check to succeed, as `v10.0-new-candidate-view+word-loader-in-notifications` doesn't make a good tag name :wink: Just give me a ping if you permanently use one of the two mentioned variants, and I set my updater accordingly.

sspanak:
> > Does Izzy ingore the ones marked as "pre-release"?
> 
> Depends. Usually I ignore them, but that's an option that can be set per-app. Let me check how it's set for yours… Yes, pre-releases are skipped. And I see you're now using sane tag and version names, so I can match tags based on them? My updater uses [semantic versioning](https://semver.org/) for that based on `versionName` or numeric comparison based on `versionCode`, depending on what tag names are used. In your case, I had to disable that and hope for the date check to succeed, as `v10.0-new-candidate-view+word-loader-in-notifications` doesn't make a good tag name wink Just give me a ping if you permanently use one of the two mentioned variants, and I set my updater accordingly.

Let's ignore pre-releases. That's for sure.

As for versioning, I don't use semantic versioning, but the versions are in the same format. I would like to keep my descriptive tags, because the source of truth for me is `git` and having such descriptions provides an easy way of finding what's in a specific version. 

However, I don't mind pushing extra tags in the format: `MAJOR.MINOR`, adding `-beta` at the end as needed, to conform with the semantic versioning format. Would that be OK for your updater? Would it matter if there is more than one tag per commit? @IzzySoft ?

sspanak (approved):

sspanak:
Oh, so you need to compare the `versionName` or `versionCode` of the `.apk` with a tag and see if they match? I was wondering why you mentioned them when talking about the releases. I suppose that's how you check for a new version, without having to download the entire `.apk`. I got it now. :slightly_smiling_face:

So, my build script needs the "v" prefix to determine whether this is a new release or not, which means I must keep at least that. 

Here is what I'll do:
- Remove any extra strings from `versionName`, as you suggested. The format will be: `11.0` or `11.0-beta`, when it is a beta. Beta versions will be pre-releases, so you don't need to worry about them.
- Tag the commit with "v" + `versionName`, in other words `v11.0` or `v11.0-beta`. 
- Add any extra tags for my convenience. They will not contain any version information, but rather will be human-readable phrases. They must be ignored by all scripts.

I believe this plan would work, except maybe when comparing versions 10.0 and 11.0, because of the non-standard `versionName` in 10.0. I'll merge this PR and proceed with the plan. In case anything goes wrong, please open an issue in my repo, so we can sort things out.

Pinging you to make sure you get a notification, @IzzySoft.

IzzySoft:
> Oh, so you need to compare the `versionName` or `versionCode` …

Yes, you've got that perfectly fine now :smiley: You can check [the code here](https://gitlab.com/IzzyOnDroid/repo/-/blob/master/lib/VersionUtils.class.php#L40). Depending on the tag naming pattern, I either compare based on the `versionCode` (very straight as that's just an integer) or the `versionName` (assuming semantic versioning structures) against what `aapt d badging`would extract from the APK I have. So I don't need to download several Gigabytes of APKs each day just to find out (my repo holds close to 1k apps as of today. Assuming each APK on average being 5 MB, that would mean 5 GB "waste" otherwise).

> So, my build script needs the "v" prefix to determine whether this is a new release or not, which means I must keep at least that.

Choose a prefix of your liking and let me know which one it is. "v" is just the most common; you could also pick "rel-" if you prefer. Just make sure to have one dedicated and always used for releases. I'd then set that up here, and my updater would ignore all other tags.

> Here is what I'll do:

That sounds perfectly fine, thanks!

> I believe this plan would work, except maybe when comparing versions 10.0 and 11.0

I'm quite confident it will work out fine (at least once the first "fine" tag was pulled). Should something go wrong I'll open another issue here as you requested. Thanks a lot!

PR #133, opened 2022-12-16
Language and dictionary fixes

sspanak:


src/io/github/sspanak/tt9/languages/Language.java
```diff
@@ -90,16 +106,17 @@ public String getDigitSequenceForWord(String word) throws InvalidLanguageCharact
...
 		for (int i = 0; i < lowerCaseWord.length(); i++) {
-			for (int key = 0; key <= 9; key++) {
-				if (getKeyCharacters(key).contains(Character.toString(lowerCaseWord.charAt(i)))) {
-					sequence.append(key);
-				}
+			char letter = lowerCaseWord.charAt(i);
+			if (!reverseCharacterMap.containsKey(letter)) {
+				throw new InvalidLanguageCharactersException(this, "Failed generating digit sequence for word: '" + word);
 			}
-		}
```
postpersonality, line 118: Not sure how hashing is optimized. But it still requires hashing even on get operation. I would suggest looking specialized maps (where key is Int) or sparse arrays: https://stackoverflow.com/questions/12626135/memory-efficient-sparse-array-in-java

postpersonality (commented):
I think hashmap is not the fastest data structure for the case.

src/io/github/sspanak/tt9/languages/Language.java
sspanak: Hm, OK, I don't have so much experience with Java, so I don't know how everything works internally. There may be better options, indeed. I'll check them out, but I think adding anything external would be an overkill. 

My main concern was optimizing the speed, not the memory usage (it is not a big deal in this case). And it definitely works much faster. With all debugging stuff disabled, the English dictionary loads in 2.6 seconds on my phone. I couldn't ask for more... :slightly_smiling_face:

postpersonality: I'm talking about speed at the first place, not memory consumption. It's just because of sparse unicode chars used it's not very efficient to use conventional array for the lookup table here, thus sparse array. There is android builtin class for that as far as I know.

sspanak: I tested with `SparseArray` and it is actually slower. In debugging mode, the English language loads in 5.7-5.8 seconds when the database is empty. Using a `HashMap` results in 5.2-5.3 seconds. I am leaving the `HashMap`.

Nevertheless, thanks for the suggestion!

PR #185, opened 2023-02-14
Add underscore to puncutation list

justbruce:
There was previously no way to access the underscore - it is a necessary character for entering URLs, amongst other things

Added right behind the dash, though where it exists doesnt matter so much as THAT it exists as a choice


sspanak:
I don't believe I have missed the underscore. Thank you very much!

Just one thing, it is not really "punctuation" in any language, so I'll move it to 0-key where the special characters are.

sspanak (approved):

PR #190, opened 2023-02-20
Hebrew support

AshiVered:
add hebrew support&stirngs, but some issues with hebrew lang icons

MAINTAINER EDIT: Merge this after #192 to avoid conflicts.

PR #199, opened 2023-02-28
Add Hebrew

DarthFlip:
Basically the additions @AshiVered made, just I added working language icons. I added the same icon for upper and lower cases just because when I didn't there was no icon at all on the top every other time it was not on t9 mode.


sspanak:
@DarthFlip, thanks, but I already took care of everything in the #190. I'm gonna have to close this one, I am sorry.

Actually, what is preventing me from merging #190 is, some extra work is needed to make the engine work properly with only one text case. And I am on it right now.

Anyway, you could help me something else. Currently, in ABC mode (Aleph-Bet, I guess :slightly_smiling_face:), it is not possible to type the final form of any letter. Is this OK? Do you guys write this way? From my perspective, as a Latin and Cyrillic user, this feels weird. I feel like I should add the final letter variants as extra options for the respective key. What do you think?

PR #217, opened 2023-03-17
Add soft key support

sspanak:
Something was wrong with #214 and #215, so I created a new branch and merged them together.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -69,6 +69,8 @@
 	<string name="key_clear" translatable="false">Clear</string>
 	<string name="key_delete" translatable="false">Delete</string>
 	<string name="key_menu" translatable="false">Menu</string>
+	<string name="key_soft_left" translatable="false">Left Select Key</string>
+	<string name="key_soft_right" translatable="false">Right Select Key</string>
 
 	<string name="char_dot" translatable="false">.</string>
 	<string name="char_newline">New Line</string>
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -60,6 +60,8 @@
 	<string name="key_clear" translatable="false">Clear</string>
 	<string name="key_delete" translatable="false">Delete</string>
 	<string name="key_menu" translatable="false">Menu</string>
+	<string name="key_soft_left" translatable="false">Left Select Key</string>
+	<string name="key_soft_right" translatable="false">Right Select Key</string>
```
sspanak:
```suggestion
	<string name="key_soft_left" translatable="false">Left Soft</string>
	<string name="key_soft_right" translatable="false">Right Soft</string>
```

sspanak:
@DarthFlip, I have renamed the keys to be consistent with the rest.

Also, I can't confirm the new hotkeys work properly, because my phone has buttons below the screen, but Android identifies them as MENU and BACK. But I suppose all is good.

DarthFlip:
I made the edits and it worked great on my phones. I attatched a picture of what my phones look like. I am trying to add the top 2 keys that are to the right/left of the numpad on top of Tlk/End and Spkr/Camera:

IMG-5360

sspanak:
> I made the edits and it worked great on my phones. I attatched a picture of what my phones look like. I am trying to add the top 2 keys that are to the right/left of the numpad on top of Tlk/End and Spkr/Camera

Got it. My point was, my phone says the same keys have different key codes.

PR #219, opened 2023-03-18
Swedish language support

KaniJonas:
Support for Swedish language


src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -271,6 +274,11 @@ protected boolean onRight(boolean repeat) {
 	protected boolean onNumber(int key, boolean hold, int repeat) {
 		String currentWord = getComposingText();
 
+		if(mInputMode.isABC()) {
```
sspanak, line 277: Perhaps, this would be a better alternative?

```suggestion
		if (delayHandler.hasCallbacks(delayAction)) {
```

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -29,6 +29,9 @@ public class TraditionalT9 extends KeyPadHandler {
 	private TextField textField;
 	private InputType inputType;
 
+	final Handler delayHandler = new Handler(Looper.getMainLooper());
+	final Runnable delayAction = () -> commitCurrentSuggestion(false);
```
sspanak, line 33: @sspanak, double check if the anonymous function works on Android 4.4.

src/io/github/sspanak/tt9/preferences/helpers/Hotkeys.java
```diff
@@ -117,7 +117,7 @@ private void generateList() {
 
 		addIfDeviceHasKey(KeyEvent.KEYCODE_BACK, R.string.key_back, false);
 		addIfDeviceHasKey(KeyEvent.KEYCODE_CLEAR, R.string.key_clear, false);
-		addIfDeviceHasKey(KeyEvent.KEYCODE_DEL, R.string.key_delete, false);
+		add(KeyEvent.KEYCODE_DEL, R.string.key_delete, false);
```
sspanak, line 120: @KaniJonas, I am not so keen on on merging this "temporary fix". It will make the option appear on phones that don't have the key.

Would it be possible for you to find another solution? I would even agree modifying `addIfDeviceHasKey()` to detect the Kyocera using `Build.MODEL`, `Build.MANUFACTURER` or some other method, and always returning `true` for that key. But whatever you choose, please, let's do it in a separate pull request.

sspanak (changes requested):

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -271,6 +274,11 @@ protected boolean onRight(boolean repeat) {
 	protected boolean onNumber(int key, boolean hold, int repeat) {
 		String currentWord = getComposingText();
 
+		if(mInputMode.isABC()) {
+			delayHandler.removeCallbacks(delayAction);
+			delayHandler.postDelayed(delayAction, 800);
```
sspanak, line 279: Let's make this an on/off setting under Keyboard -> ABC Mode. Personally, I find it confusing when the keyboards auto-accept letters or words, so I would like to keep it turned off.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
sspanak: It appears to be fine.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
KaniJonas: Hello, I will have to apologize, I am quite new to GitHub and didn't realize these changes would get included in the pull request for the Swedish language.

My intention was to only use the changes for Traditional T9 and Hotkeys in my own fork. Possibly contributing with the delay function later on with on/off option and adjustable delay.

sspanak: But I do want to include this feature for other people to use it. It's nice that you shared it. :slightly_smiling_face: 

~~You almost got it perfect, it's just missing the toggle switch. If you don't mind, I can add the setting in your branch, when I get free time in the following days.~~

After a more careful review, I think it needs a bit more work, but it is a good start and I can include most of the code later.

sspanak:
I have extracted only the code related to Swedish language in #223 and merged it. I will now close this PR, because it has become obsolete.

Feel free to open a new one only for the ABC auto accept feature or the fix for Kyocera.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
KaniJonas: Thanks, that sounds wonderful if you have the time! :) If you don't, I can probably figure most of it out myself when I get the chance.
I can't in words tell you how much I love this app and all the time you put down. I've been suffering with Nokia's buggy horrible T9 for years.

PR #223, opened 2023-03-21
Swedish language

sspanak:
Extracted only the language update from #219.


src/io/github/sspanak/tt9/languages/LanguageCollection.java
```diff
@@ -33,10 +33,11 @@ private LanguageCollection() {
 			Norwegian.class,
 			Russian.class,
 			Spanish.class,
+			Swedish.class,
 			Ukrainian.class
 		);
 
-		// initialize the language objects from the class list above
+		// initialize the language objects from the class list above.
```
sspanak:
```suggestion
		// initialize the language objects from the class list above
```

PR #230, opened 2023-03-28
Status bar

sspanak:


src/io/github/sspanak/tt9/ui/bottom/SuggestionsBar.java
```diff
@@ -159,14 +163,30 @@ public void scrollToSuggestion(int increment) {
...
-		int backgroundColor = darkEnabled ? R.color.dark_candidate_background : R.color.candidate_background;
+
 		int defaultColor = darkEnabled ? R.color.dark_candidate_color : R.color.candidate_color;
 		int highlightColor = darkEnabled ? R.color.dark_candidate_selected : R.color.candidate_selected;
 
-		mView.setBackgroundColor(ContextCompat.getColor(context, backgroundColor));
 		mSuggestionsAdapter.setColorDefault(ContextCompat.getColor(context, defaultColor));
 		mSuggestionsAdapter.setColorHighlight(ContextCompat.getColor(context, highlightColor));
+
+		setBackground();
```
sspanak: Determining and setting the color on every suggestion change is suboptimal. Find a way of doing it only when the `suggestions.size()` becomes 0 or different than 0.

Also, the status flashes for a split second when new suggestions are loading. Fix this.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -521,7 +528,7 @@ else if (mInputMode.isABC() && mInputMode.getTextCase() == InputMode.CASE_LOWER
 		settings.saveInputMode(mInputMode.getId());
 		settings.saveTextCase(mInputMode.getTextCase());
 
-		UI.updateStatusIcon(this, mLanguage, mInputMode);
+		statusBar.setText(mInputMode.toString());
```
sspanak, line 531: Use a toast notification when in `EDITING_NOSHOW` mode. Create a `UI.updateStatus()` or similarly-named function that chooses the appropriate way of notifying the user.

PR #231, opened 2023-03-28
Soft keyboard numpad

nebkrid:
continuing #188 

> If I haven't answered or have forgotten about some of your questions, just copy them and I'll do there.
Now, regarding the long press to type a number, this may not be strictly necessary on touchscreen keyboard. From what I saw, all other keyboards, require manually switching to numeric mode to type numbers, which means we are all good in this regard.

I think when we skip the long-press-to-type-number-function all discussed points are now solved - except from the restructuring which you have in mind.


sspanak:
@nebkrid, I ended up refactoring quite a lot of code. I had squash everything into a single commit for easier conflict solving and updating with `master`. This means, if you want to use the branch again, you must pull with: 
```
$ git pull --rebase
```

sspanak (approved):

Issue #53, opened 2022-09-11
Consider using a Trie for dictionary storage


sspanak:
Instead of the flat storage in SQL, try using a Trie. It may be much faster and use less space. When switching, ensure the suggestions still have priorities.

Educational info:
- [Succint Data Structures](http://stevehanov.ca/blog/index.php?id=120)
- [Compressing Dictionaries](http://stevehanov.ca/blog/index.php?id=115)

[A DAWG](https://en.wikipedia.org/wiki/Deterministic_acyclic_finite_state_automaton) may be even a better option, because it allows sharing common intermediate nodes. For example:
```
t | o | p
  | a |
  | i |
```
This may be useful for compressing the assets. Here is a nice [implementation in Python](https://github.com/turekj/msc/tree/master/Scrabble/modules/dictionary/dawg).

Clam-:
You may find that SQLite is quite optimized at searching and storing due to the underlying tree nature of its internals.
Historically the very first version of the app I attempted to use a trie-like data structure. I wasn't very successful, since I wasn't sure how to optimally store it in memory vs disk and ended up consuming all memory trying to load a dictionary haha.
SQLite does all that messy work for you, however even in saying all that I would be very interested to see what you end up creating.
Good luck!

sspanak:
> You may find that SQLite is quite optimized at searching and storing due to the underlying tree nature of its internals.

This is not really something absolutely necessary. SQLite works just fine. Even in debug mode it loads the suggestion list in less than 10 ms, so there is no need to replace it.

I was not familiar with this data structure. I found about it accidentally, while I was working on another tree in my real job. The demo in the first link seemed quite impressive, so I though I could experiment with it, when I have enough free time. But of course, I immediatelly realized that a relational database is not an optimal storage for this data structure and I knew I had to find a better one.

Anyway, my interest in this could be classified as purely scientific.

> Historically the very first version of the app I attempted to use a trie-like data structure. I wasn't very successful, since I wasn't sure how to optimally store it in memory vs disk and ended up consuming all memory trying to load a dictionary

That's strange. According to many sources this _is_ the data structure for dictionaries. It's the most optimal one and it has been used in early phones that had very little processing power and memory. Who knows... you may have not been successful, because of the way Java works... I may find one day!

sspanak:
This is an interesting academic problem, but a more realistic approach is replacing the database as described in #366. If performance issues still persist, this can be reopened.

PR #234, opened 2023-04-01
Add Polish language

rjolina:
i added polish language according to your instructions.
i am aware that the dictionary file is 10 times larger than other dictionaries, but the Polish language has a lot of declensions and I do not know if it is possible to "trim" it somehow. This is also my first pull request on github, please be understanding :)


sspanak:
Thank you for your contribution. It is a great dictionary, but indeed, 60 Mb is an overkill.

I was thinking of breaking the dictionaries into a list of word roots, prefixes and suffixes, hoping to make it possible to use much more complete dictionaries. Or even [using a Trie](https://github.com/sspanak/tt9/issues/53) for that matter. I guess this would be a great starting point for experiments. :slightly_smiling_face: 

So let's hold on with this PR until we find a better solution.

sspanak:
> I believe that polish grammar is quite similar to ukrainian/bulgarian/russian

Not at all... :slightly_smiling_face: Bulgarian has no declensions and I still feel I am missing words every now and then. I suppose the optimum would be at around 350k words.

Ukrainian grammar resembles Polish and the dictionnary may be ok-ish, but Russian probably feels quite incomplete. I think less than 200k is fine for a language like English, where words do not change much. I do use it and I am speaking from personal experience.

I don't mind adding one of the Google dictionaries, but you will probably find yourselves adding new words all the time. Of course, I may be wrong. I'm fine with testing this option, if you want to. I would appreciate your feedback, too. Or, if you could find another word list of, say, 350k-500k words, we can go for that.

But we can not use this dictionary now, it is too big. Some lower spec phones may crash trying to load it.

Either way, as I said in my previous comment, I want to find a way of improving the dictionary storage mechanism and expanding the dictionaries for better experience.

samurex:
> if you could find another word list of, say, 350k-500k words, we can go for that.

This one looks good https://mirrors.tuna.tsinghua.edu.cn/ctan/systems/windows/winedt/dict/pl.zip, 500k words. What do you guys think ?

sspanak:
> > if you could find another word list of, say, 350k-500k words, we can go for that.
> 
> This one looks good https://mirrors.tuna.tsinghua.edu.cn/ctan/systems/windows/winedt/dict/pl.zip, 450k words. What do you think ? Just has fancy windows cp1250 encoding that needs to be converted to utf-8

It should be OK. And the encoding is fine, I may even keep it to save some disk space.

rjolina:
> > > if you could find another word list of, say, 350k-500k words, we can go for that.
> > 
> > 
> > This one looks good https://mirrors.tuna.tsinghua.edu.cn/ctan/systems/windows/winedt/dict/pl.zip, 450k words. What do you think ? Just has fancy windows cp1250 encoding that needs to be converted to utf-8
> 
> It should be OK. And the encoding is fine, I may even keep it to save some disk space.

This dictionary looks OK also for me. Should I make a new pull request or somehow modify this one?

samurex:
I'm not a contributor of this project, just want to have good T9 keyboard on my phone :) How about adding word frequencies from [openboard](https://github.com/openboard-team/openboard/blob/master/dictionaries/pl_wordlist.combined.gz) ?

sspanak:
> I'm not a contributor of this project, just want to have good T9 keyboard on my phone :) How about adding word frequencies from [openboard](https://github.com/openboard-team/openboard/blob/master/dictionaries/pl_wordlist.combined.gz) ?

This is a must in order to have good experience. If you have `nodejs` installed, you could use `scripts/inject-dictionary-frequencies.js` for that. If not, I will do so before merging.

src/io/github/sspanak/tt9/languages/definitions/Polish.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Locale;
+
+public class Polish extends English {
+	public Polish() {
+		super();
+
+		locale = Locale.POLISH;
+		dictionaryFile = "pl-utf8.txt";
+
+		isPunctuationPartOfWords = true;
```
sspanak: This must be off. Polish words don't contain apostrophes like "don't".

```suggestion
```

src/io/github/sspanak/tt9/languages/definitions/Polish.java
```diff
@@ -0,0 +1,22 @@
...
+
+		characterMap.get(2).addAll("ą", "ć"));
+		characterMap.get(3).add(Arrays.asList("ę");
+		characterMap.get(5).add(Arrays.asList("ł");
+		characterMap.get(6).addAll(Arrays.asList("ó", "ń"));
+		characterMap.get(7).add(Arrays.asList("ś");
+		characterMap.get(9).addAll(Arrays.asList("ź", "ż"));
```
sspanak:
```suggestion
		characterMap.get(2).addAll("ą", "ć"));
		characterMap.get(3).add("ę");
		characterMap.get(5).add("ł");
		characterMap.get(6).addAll(Arrays.asList("ó", "ń"));
		characterMap.get(7).add("ś");
		characterMap.get(9).addAll(Arrays.asList("ź", "ż"));
```

docs/dictionaries/plWordListReadme.txt
```diff
@@ -0,0 +1,4 @@
+Polish wordlist by: rjolina
```
sspanak: This information is incorrect. Please, replace it with the original readme from the `.zip` file, excluding the encoding and win 95 stuff.

As for the license, nothing is explicitly stated by the author, so I suppose, we could consider the word list is compiled from public domain sources. But it is definitely not GPL 2.0.

sspanak (changes requested):

src/io/github/sspanak/tt9/languages/definitions/Polish.java
```diff
@@ -0,0 +1,22 @@
+package io.github.sspanak.tt9.languages.definitions;
+
+import java.util.Arrays;
+import java.util.Locale;
+
+public class Polish extends English {
+	public Polish() {
+		super();
+
+		locale = Locale.POLISH;
```
sspanak: I am not sure this is a valid locale in Java. Please, make sure to correct everything that causes the build to fail.

src/io/github/sspanak/tt9/languages/definitions/Polish.java
mikep-dev: Do compound words count? E.g. "czarno-biały" means "black and white" and includes a dash.

sspanak: There are no such words in the word list, so there is no need to enable the option.

Either way, it was a hack for Dutch language, where there are words that start with an apostrophe. It wasn't a great idea, but at the time, I couldn't come up with anything better. But now, I am planning to get rid of it, so don't worry too much about it.

sspanak:
I've noticed some suposingly nonsense words, such as: "Md", "np", "xi", "xv", "iii", "xxx" and so on. 

I would appreciate someone who speaks the language to review clean up the two- and three-letter words. Otherwise, I'll just remove anything that looks weird.

mikep-dev:
Hey @sspanak, I tried out that dictionary and it's kind of wonky. It has some really strange words and is missing a lot of basic ones. I'm going to work on fixing it up or finding a better one, but it might take me a bit since I don't have much spare time lately.

sspanak:
So, I've cleaned up the dictionary and tried it out by typing a couple of sample sentences from Wikipedia. It is mostly fine, even though I found one word was missing. I suppose the experience is good enough, so I will merge the PR.

I suppose the optimal word count for Polish would be around 800k-1M words. @rjolina, @mikep-dev, @samurex, if you could find such a dictionary, feel free to open a new pull request.

sspanak (approved):

rjolina:
Thank you guys for your cooperation! Can't wait for the release with my native language 🥰

PR #238, opened 2023-04-15
Allow switching languages midword

flaviozavan:
Allows switching languages midword.
When in predictive mode, a new match is searched for based on the previously entered keys, allowing for quick corrections between languages. When in ABC mode, the current letter is not committed, and the letter list is updated.

Since the status bar is no longer visible when typing, immediate feedback is shown by using a toast.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -353,7 +353,6 @@ public class TraditionalT9 extends KeyPadHandler {
 			mInputMode.clearWordStem();
 			clearSuggestions();
 			getSuggestions();
-			resetKeyRepeat();
 
 			statusBar.setText(mInputMode.toString());
 			mainView.render();
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -349,14 +349,17 @@ public class TraditionalT9 extends KeyPadHandler {
 
 	public boolean onKeyNextLanguage() {
 		if (nextLang()) {
-			commitCurrentSuggestion(false);
 			mInputMode.changeLanguage(mLanguage);
-			mInputMode.reset();
+			mInputMode.clearWordStem();
+			getSuggestions();
 			resetKeyRepeat();
-			clearSuggestions();
+
 			statusBar.setText(mInputMode.toString());
 			mainView.render();
 			forceShowWindowIfHidden();
+			if (!isSuggestionViewHidden()) {
+				UI.toastLang(getMainContext(), mInputMode.toString());
+			}
 
 			return true;
 		}
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -351,20 +351,19 @@ public class TraditionalT9 extends KeyPadHandler {
 		if (nextLang()) {
 			mInputMode.changeLanguage(mLanguage);
 			mInputMode.clearWordStem();
+			clearSuggestions();
 			getSuggestions();
 			resetKeyRepeat();
 
 			statusBar.setText(mInputMode.toString());
 			mainView.render();
 			forceShowWindowIfHidden();
-			if (!isSuggestionViewHidden()) {
+			if (mInputMode.isABC() && !isSuggestionViewHidden()) {
 				UI.toastLang(getMainContext(), mInputMode.toString());
 			}
-
-			return true;
 		}
 
-		return false;
+		return true;
 	}
 
 
@@ -574,9 +573,8 @@ public class TraditionalT9 extends KeyPadHandler {
 			return false;
 		}
 
-		// when only one language is enabled, just acknowledge the key was pressed
 		if (mEnabledLanguages.size() < 2) {
-			return true;
+			return false;
 		}
 
 		// select the next language
--- a/src/io/github/sspanak/tt9/ime/modes/ModeABC.java
+++ b/src/io/github/sspanak/tt9/ime/modes/ModeABC.java
@@ -4,8 +4,6 @@ import androidx.annotation.NonNull;
 
 import io.github.sspanak.tt9.languages.Language;
 
-import java.util.ArrayList;
-
 public class ModeABC extends InputMode {
 	public int getId() { return MODE_ABC; }
 
@@ -19,16 +17,15 @@ public class ModeABC extends InputMode {
 
 	@Override
 	public boolean onNumber(int number, boolean hold, int repeat) {
-		digit = number;
 		if (hold) {
 			reset();
-			digit = -1;
-			suggestions.add(String.valueOf(number));
 			autoAcceptTimeout = 0;
+			suggestions.add(String.valueOf(number));
 		} else if (repeat > 0) {
 			shouldSelectNextLetter = true;
 		} else {
 			reset();
+			digit = number;
 			suggestions.addAll(language.getKeyCharacters(number));
 		}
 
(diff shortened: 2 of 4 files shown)


src/io/github/sspanak/tt9/ui/UI.java
```diff
@@ -46,4 +48,13 @@ public static void toastLong(Context context, int resourceId) {
 	public static void toastLong(Context context, CharSequence msg) {
 		Toast.makeText(context, msg, Toast.LENGTH_LONG).show();
 	}
+
+	public static void toastLang(Context context, CharSequence msg) {
+		if (langToast != null) {
+			langToast.cancel();
+		}
+
+		langToast = Toast.makeText(context, msg, Toast.LENGTH_SHORT);
```
sspanak, line 57:
```suggestion
		} else {

		langToast = Toast.makeText(context, msg, Toast.LENGTH_SHORT);
```

sspanak, line 52: Is there any need to re-create the Toast every time? Isn't this more efficient:
```suggestion
		if (langToast == null) {
			langToast = Toast.makeText(context, msg, Toast.LENGTH_SHORT);
		} else {
			langToast.cancel();
		}

```
?

Btw, I've been trying to get rid of these toasts, but they keep coming back... :disappointed:

src/io/github/sspanak/tt9/ime/modes/ModeABC.java
```diff
@@ -10,6 +10,7 @@ public class ModeABC extends InputMode {
 	public int getId() { return MODE_ABC; }
 
 	private boolean shouldSelectNextLetter = false;
+	private int lastKey = -1;
```
sspanak: Let's rename this for consistency with `ModePredictive`
```suggestion
	private int digit = -1;
```

src/io/github/sspanak/tt9/ui/UI.java
flaviozavan: I tried a few different solutions here. If we reuse the toast, sometimes they don't appear (about 10% of the time), even when they are explicitly cancelled beforehand.

> Btw, I've been trying to get rid of these toasts, but they keep coming back... disappointed

I gathered that much from the recent commits, but couldn't come up with a better idea :grimacing:

src/io/github/sspanak/tt9/ime/modes/ModeABC.java
flaviozavan: sounds reasonable

src/io/github/sspanak/tt9/ui/UI.java
sspanak: Yeah, well... let's keep it. I can't come up with anything better either.

sspanak:
@flaviozavan, I've tried out this feature, but it doesn't work well in ABC mode - the letters do not update properly and Backspace acts weirdly, typing characters instead of erasing them.

And in Predictive mode, it actually combines the starting of the word in the first language with the ending in another, instead of replacing the entire word with the equivalent in the second language. But combining two different words makes little sense, especially if the languages are written in different alphabets.

At first, the idea looked awesome to me, but now I am thinking of maybe declining the PR... it may not be that good after all.

flaviozavan:
The ABC mode behaviors you described sound like bugs I thought I had fixed, but there might be edge cases. And what you see in predictive mode is not intended and should be a bug, I've been using the feature all week and the entire word gets replaced for me. Maybe these issues are related to different alphabets. I'll take a look and fix when I have some free time.

flaviozavan:
Rebased and introduced a new commit.

It seems the backspace and ABC mode issues were likely due to bad conflict resolution, as I wasn't able to reproduce them. Happy to take another look if you can.

As for the mixing of alphabets, it was a side-effect of there being no words for a given key combination. This was present with languages of similar alphabets, it just wasn't as jarring. The new commit changes the Predictions class to take into account language changes, replacing impossible characters

src/io/github/sspanak/tt9/ime/modes/helpers/Predictions.java
```diff
@@ -70,6 +76,23 @@ public Predictions setStem(String stem) {
...
+					newInputWord += this.inputWord.charAt(i);
+				}
+				else {
+					newInputWord += keyChars.get(0);
+				}
+			}
+			this.inputWord = newInputWord;
+			inputWordMightBeForeign = false;
+		}
+
```
sspanak:
```suggestion
```

This is not necessary. 
1. In many cases it will generate a nonsense word, especially if the two languages use different alphabets.
2. Calling `getSuggestions()` will result in erasing the generated word, and populating the text field with the first suggestion from the database.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -349,14 +349,17 @@ public boolean onKeyAddWord() {
...
-			mInputMode.reset();
+			mInputMode.clearWordStem();
+			getSuggestions();
 			resetKeyRepeat();
-			clearSuggestions();
+
 			statusBar.setText(mInputMode.toString());
 			mainView.render();
 			forceShowWindowIfHidden();
+			if (!isSuggestionViewHidden()) {
```
sspanak:
```suggestion
			if (mInputMode.isABC() && !isSuggestionViewHidden()) {
```

In Predictive mode it will be obvious the language has changed, because the word list would become entirely different. So, let's reduce the toast messages to the bare minimum.

sspanak (changes requested):

sspanak:
@flaviozavan, it is looking good this time. There are just two more small things and this will be ready to go. 

If you don't have time today or tomorrow, do you mind if I update and merge the pull request?

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -349,14 +349,17 @@ public boolean onKeyAddWord() {
 
 	public boolean onKeyNextLanguage() {
 		if (nextLang()) {
-			commitCurrentSuggestion(false);
 			mInputMode.changeLanguage(mLanguage);
-			mInputMode.reset();
+			mInputMode.clearWordStem();
+			getSuggestions();
 			resetKeyRepeat();
```
sspanak:
```suggestion
			clearSuggestions();
			getSuggestions();
			resetKeyRepeat();
```

`clearSuggestions()` + `getSuggestions()` is the equivalent of what you tried to do in `setInputWord()`. Let's do it this way.

src/io/github/sspanak/tt9/ime/modes/helpers/Predictions.java
```diff
@@ -22,6 +22,7 @@ public class Predictions {
 	private boolean isStemFuzzy;
 	private String stem;
 	private String inputWord;
+	private boolean inputWordMightBeForeign;
```
sspanak:
```suggestion
```

flaviozavan:
@sspanak Hey. The suggestions look alright.

getSuggestions: Your idea seems better
About no toasts on ABC: There are languages that share alphabets (e.g. Finnish and Swedish), but the impact of having the wrong one in ABC mode is almost zero. So that seems fine as well.

I don't have the bandwidth to work on this before the weekend. Feel free to change and merge

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -349,19 +349,20 @@ public boolean onKeyAddWord() {
 
 	public boolean onKeyNextLanguage() {
 		if (nextLang()) {
-			commitCurrentSuggestion(false);
 			mInputMode.changeLanguage(mLanguage);
-			mInputMode.reset();
-			resetKeyRepeat();
```
sspanak, line 355: ~~This is still causing problems. Try changing to 123 mode, pressing any number, then changing the language. It may cause other side effects, so key repeat must be reset at some point, but without breaking ABC mode.~~

~~Also,~~ in Predictive mode, when there are no more words, the suggestion list gets populated with letters from the penultimate keypress.

This PR will definitely need some more work before merging.

sspanak (changes requested):

PR #250, opened 2023-05-04
Speed optimizations

sspanak:


src/io/github/sspanak/tt9/db/room/Word.java
```diff
@@ -8,7 +8,8 @@
 @Entity(
 	indices = {
 		@Index(value = {"lang", "word"}, unique = true),
-		@Index(value = {"lang", "seq", "freq"})
+		@Index(value = {"lang", "seq", "freq"}, orders = {Index.Order.ASC, Index.Order.ASC, Index.Order.DESC}),
+		@Index(value = {"lang", "len", "seq"}, orders = {Index.Order.ASC, Index.Order.ASC, Index.Order.ASC})
```
sspanak, line 12: 1. Use the "lang_seq_freq" index when sequence length > 2 and "lang_len_seq" when length <= 2. A `@RawQuery` will be required, because RoomDB does not support `SELECT xxx FROM tbl INDEXED BY index_name`

2. Get rid of these (if possible). Instead create Room functions for dynamically creating and dropping the indexes when bulk inserting in `DictionaryLoader`.

src/io/github/sspanak/tt9/db/DictionaryLoader.java
sspanak, line 1: Before bulk inserting, drop all indexes, then recreate them. The operation will be much faster.

PR #259, opened 2023-05-16
upgrade gradle 7.4.2 -> 8.0.1

sspanak:


sspanak:
Must figure out what is wrong with upgrading `androidx.core:core`

PR #265, opened 2023-05-26
mEditing is no more

sspanak:


src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -556,8 +569,9 @@ else if (mInputMode.isABC() && mInputMode.getTextCase() == InputMode.CASE_LOWER
 		} else {
 			int modeIndex = (allowedInputModes.indexOf(mInputMode.getId()) + 1) % allowedInputModes.size();
 			mInputMode = InputMode.getInstance(settings, mLanguage, allowedInputModes.get(modeIndex));
+			mInputMode.setTextFieldCase(textField.determineTextCase(inputType));
+			mInputMode.determineNextWordTextCase(textField.isThereText(), textField.getTextBeforeCursor());
```
sspanak, line 581: 1. Use the on-screen keyboard
2. Switch from ABC to Predictive Mode.

The "inscribed" letters appear in uppercase, but the actual text case is not uppercase. Fix this!

sspanak: Fixed.

PR #267, opened 2023-05-27
Support for "I" in the English dictionary.

alexknop:
Problem:
Since "I" is a word in the English language, there was a conflict where "I","g","h" were the only suggestions in the middle of a sentence when in Predictive Mode. For someone building out a word manually in Predictive mode, they would need a lowercase i.

Solution:
1. Removed check when loading a dictionary that is seeing if the uppercase version of a letter is already there. This is preventing "i" from being in the dictionary, thus it wouldn't show in Predictive mode.
2. Removed check that could prevent "I" from being a suggestion because it would convert it to a lowercase i and compare it to "i" which may already be a suggestion, and thus exclude it from the final list of suggestions.
3. Added logic that in an uppercase list of suggestions when the digit sequence in Predictive mode is 4 (G H I I), removes duplicates so that the list of suggestions are just (G H I).
4. Fixed bug that was still calling getSuggestions() after deleting the first character of a sequence which always returned 0 suggestions (which was expected). Calling this was unneeded, so we are now properly returning False on onBackspace().

### DIFF
--- a/assets/en-utf8.csv
+++ b/assets/en-utf8.csv
@@ -1,4 +1,3 @@
-I
 abs	62
 cia	118
 co
--- a/assets/en-utf8.csv
+++ b/assets/en-utf8.csv
@@ -1,3 +1,4 @@
+I
 abs	62
 cia	118
 co
--- a/src/io/github/sspanak/tt9/db/DictionaryLoader.java
+++ b/src/io/github/sspanak/tt9/db/DictionaryLoader.java
@@ -202,9 +202,9 @@ public class DictionaryLoader {
 					continue;
 				}
 
-				if (DictionaryDb.doesWordExistSync(language, langChar.toUpperCase(language.getLocale()))) {
-					continue;
-				}
+//				if (DictionaryDb.doesWordExistSync(language, langChar.toUpperCase(language.getLocale()))) {
+//					continue;
+//				}
 
 				Word word = new Word();
 				word.langId = language.getId();
--- a/src/io/github/sspanak/tt9/db/DictionaryLoader.java
+++ b/src/io/github/sspanak/tt9/db/DictionaryLoader.java
@@ -202,10 +202,6 @@ public class DictionaryLoader {
 					continue;
 				}
 
-//				if (DictionaryDb.doesWordExistSync(language, langChar.toUpperCase(language.getLocale()))) {
-//					continue;
-//				}
-
 				Word word = new Word();
 				word.langId = language.getId();
 				word.frequency = 0;
--- a/src/io/github/sspanak/tt9/ime/modes/InputMode.java
+++ b/src/io/github/sspanak/tt9/ime/modes/InputMode.java
@@ -3,6 +3,8 @@ package io.github.sspanak.tt9.ime.modes;
 import androidx.annotation.NonNull;
 
 import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import io.github.sspanak.tt9.Logger;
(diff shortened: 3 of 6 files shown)


sspanak:
@alexknop, your code changes will mess up so many things, I simply can't accept them.

There is a simpler of achieving what you want. You may want to try the following:

1. Go to `ime/modes/helpers/AutoTextCase.java`.
2. At the beginning, add a new regex: `private final Pattern nextToWordRegex = Pattern.compile("\\b$");`
3. At the of `determineNextWordTextCase()` replace `return InputMode.CASE_DICTIONARY;` with:
```
		if (nextToWordRegex.matcher(textBeforeCursor).find()) {
			return InputMode.CASE_LOWER;
		}

		return InputMode.CASE_DICTIONARY;
```

This will enforce lowercase when the cursor is after or in the middle of a word.

alexknop:
I will look into this solution. I have a couple of questions on some things which I've added comments to

sspanak:
> I will look into this solution. I have a couple of questions on some things which I've added comments to

I see no comments but this one...

Anyway, I already did it in #272 and it is looking good. I'll test a bit more during the weekend and then I'll merge it.

sspanak:
> Basically, I am questioning the purpose of line 205 in DictionaryLoader.Java.:
> this looks like it is checking whether A,B,C,D,etc exist in the dictionary already..
> If so, is this only there to look out for "i" and "I"?

Yes, it is there to ensure only one variant of the letter is in the database. For performance and disk space reasons, the number of words (and letters) must be reduced to minimum. And this is OK with all languages, but English, where capital "I" can appear in the middle of a sentence.

There is an algorithm for determining the text case in the file I mentioned above. Basically, it tries to guess whether it is the beginning of a new sentence and capitalizes the letters or it preserves uppercase, if the user has selected it. However, words that must always be capitalized, such as proper names and "I" in English are stored capitalized in the database. That's the reason this hack is necessary.

However, it has spread around the code, so I am going to contain it, when I get to #276.

> And then in ModePredictive.java, shouldn't it return false after line 58?
No, returning true is correct. Returning false means the Mode hasn't processed the key press in anyway and it just passes it through to the next handler (in the next application or the system). On line 58, we have just deleted the last character of the word and we have cleared the word stem. In other words, we have processed the key press, so we must return true.

Anyway, your approach is not in line with the standard I am trying to apply. There are sort of "plugins" for the automatic text processing and the text case is handled in `ime/modes/helpers/AutoTextCase.java`. I would like to keep the code organized this way.

But, as I said, you don't need to worry about this anymore. I have already done the same in a simpler way in #272.

alexknop:
> No, returning true is correct. Returning false means the Mode hasn't processed the key press in anyway and it just passes it through to the next handler (in the next application or the system). On line 58, we have just deleted the last character of the word and we have cleared the word stem. In other words, we have processed the key press, so we must return true.

I was looking at the usage for onBackspace() and if it returns true, then getSuggestions() will be called. however if you just deleted the last character of a word, then there won't be any suggestions anyway, correct?
In my testing, I did see it calling getSuggestions() in this instance and it seemed unnecessary because there would only be suggestions if at least a single character of a word existed.

sspanak:
> I was looking at the usage for onBackspace() and if it returns true, then getSuggestions() will be called. however if you just deleted the last character of a word, then there won't be any suggestions anyway, correct?
> In my testing, I did see it calling getSuggestions() in this instance and it seemed unnecessary because there would only be suggestions if at least a single character of a word existed.

It must be called one last time, to return 0 suggestions and which clears the list.

PR #269, opened 2023-05-30
Auto accept in abc mode

sspanak:


sspanak:
Holding a key to type a number in Predictive mode fails. Fix this.

Issue #187, opened 2023-02-16
Basic Punctuation Suggestions Not Offered On '1'-Key Press After Commonly Contracted Words


levone1:
When pressing the '1' key while a word suggestion is highlighted, there are a number of suggestions presented which show common punctuation, if you want to add a period or comma, etc., after the word.  This behavior is missing if the word that's highlighted is one which is commonly contracted, like "you", (you'll, you're, etc.), or "do", (don't).  In those cases, only the common contraction suggestions are shown, not the punctuation suggestions.

Thanks

levone1:
Maybe best if the contractions were just part of the regular word suggestions, like if you type "can", then "can't" would just show in the regular suggestion list, and pressing '1' would only be for punctuation...?

AF1512:
That might be better, so you can also type "dont" and get "don't" (instead of "foot").

louisabraham:
I was going to open this issue. I think indeed that there are two ways to handle this: storing the contracted words with or without the key 1. I think that storing them with 1 might make the dictionnary more efficient since it adds information at input time.

sspanak:
> That might be better, so you can also type "dont" and get "don't" (instead of "foot").

You are pressing the wrong keys to type "don't".  Instead of "3668", use "3661" to get the correct suggestion. :slightly_smiling_face: 

I realize it's a habit caused by using touchscreen keyboards and they do offer such shortcuts for a reason, because the apostrophe is difficult to access, unlike in T9; and because of swipe typing. But typing using physical keys is different and happens under different set of rules.

So, I will fix the problem with the missing suggestions:
- Typing `you'`, should bring up all variants that end with a punctuation mark. There is definitely a bug.
- Typing `you`, should bring up `you're` or `you'd`, but sometimes they are missing. There may be something wrong or it may be because of the learned word priority on my phone. Either way, I'll check.

But I will not:
- Add "shortcut" words, for example: `dont` that auto-converts to `don't`. This will not reduce the key presses, therefore it has no value for T9 typing.
- Exclude 1-key from typing contractions. In Dutch language, for example, there are words that start with an apostrophe: `'s-Graveland` or `'t`. If 1-key types only punctuation, they will never appear as suggestions and it would be impossible to type them (except letter by letter). It will also cause other techincal problems, so it is completely out of the question.

louisabraham:
> You are pressing the wrong keys to type "don't". Instead of "3668", use "3661" to get the correct suggestion. 🙂

[...]

PR #277, opened 2023-06-02
added missing contractions

alexknop:
these are all missing English contractions from the dictionary. They are difficult to add because of the apostrophe. It also just makes the built-in dictionary more extensive for users.

### DIFF
--- a/assets/en-utf8.csv
+++ b/assets/en-utf8.csv
@@ -14,6 +14,44 @@ gi	92
 gnp	79
 gop	95
 god	124
+he's
+she's
+who's
+what're
+what'll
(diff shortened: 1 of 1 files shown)


sspanak:
Thanks for your contribution, but all these are going to be deleted in #187. I have come up with a better way of doing all this and such apostrophe pairs no longer need to be stored in the database. There could be exceptions, but I would like to get rid of them in general. Typing punctuation will also become much easier. I know it is a bit slow and annoying in English, Dutch and French.

So, please, have some patience and wait for #187. And sorry for declining a PR of yours again.

PR #278, opened 2023-06-02
remove logic for numbers, instead using includeDigit

alexknop:
Simplified logic to use the included optional boolean of includeDigit. This way we do not have to check for it in the if statement.


sspanak:
Deleting some legacy code is always a nice thing, thanks! :smile:

PR #280, opened 2023-06-05
Fix small regressions

sspanak:


src/io/github/sspanak/tt9/ime/modes/helpers/AutoSpace.java
```diff
@@ -7,6 +7,7 @@
 import io.github.sspanak.tt9.preferences.SettingsStore;
 
 public class AutoSpace {
+	private final Pattern isLetter = Pattern.compile("^\\p{L}+$");
```
sspanak: This does not properly work for "Да" -> "Дра"

sspanak: Fixed.

PR #289, opened 2023-06-16
Easier contraction typing

### LINKED ISSUE #187: Basic Punctuation Suggestions Not Offered On '1'-Key Press After Commonly Contracted Words

### LINKED ISSUE #284: Automatic space is wrongly inserted in some cases
Scenario 1 (auto-space after whitespace)
1. Type any word in any language (more than 1 letter)
2. Press "0"
3. Select space or new line
4. Press "OK"

The selected character is inserted, but also auto-space option kicks in and insert one more space, which is wrong. Space must be inserted only after the characters listed in: "AutoSpace.shouldAddAfterPunctuation()"

Scenario 2: (auto-space before a number)
1. Type: "word -"
2. Hold a key to type a number

An extra space is inserted before the number, when it mustn't

Scenario 3: (auto-space in compound words)
1. Type: "word"
2. Type "-"
3. Press "OK"

A space is inserted after "-", but there must be no space except after the strings listed in `AutoSpace.shouldAddAfterPunctuation()`.

To simplify the text analysis and get rid of the `nextKey` hack, `autoCorrectSpace()` must be called after receiving the new suggestions list. However, this also depends on auto-accepting, which is still in `onNumber()`. So do this after #187, when auto-accepting will be moved after receiving the suggestions, too.

sspanak:
Fixes #284
Fixes #187


assets/nl-utf8.csv
```diff
@@ -1,14 +1,19 @@
-'s-Graveland	94
```
sspanak, line 1: Create a migration that deletes these words

sspanak:
The priority of `'s` and `'re` is not properly updated.

PR #294, opened 2023-06-30
Update strings.xml (spanish)

gabriel-tandil:
complete Spanish language strings


sspanak:
Many thanks! Language contributions are especially valuable - only a native speaker (or a very fluent one) can do them properly.

PR #309, opened 2023-07-20
added DONE to allow selecting suggested names in messaging

alexknop:
Problem:
When typing a new text message and in the "To" box using OK to select a suggested contact name, I noticed the app was not handling an Action ID of 6 (IME_ACTION_DONE).
Solution:
By adding code for this Action ID, this now allows the Center/OK key to select a contact properly from the drop-down.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/helpers/TextField.java
+++ b/src/io/github/sspanak/tt9/ime/helpers/TextField.java
@@ -326,20 +326,18 @@ public class TextField {
 			return EditorInfo.IME_ACTION_NONE;
 		}
 
+		int action;
 		if (field.actionId > 0) {
-			return field.actionId; // custom action, defined by the connected app
+			action = field.actionId; // custom action, defined by the connected app
+		} else {
+			action = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);
 		}
-
-		int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);
-		switch (standardAction) {
-			case EditorInfo.IME_ACTION_GO:
-			case EditorInfo.IME_ACTION_NEXT:
-			case EditorInfo.IME_ACTION_PREVIOUS:
-			case EditorInfo.IME_ACTION_SEARCH:
-			case EditorInfo.IME_ACTION_SEND:
-				return standardAction;
-			default:
+		switch (action) {
+			case EditorInfo.IME_ACTION_UNSPECIFIED:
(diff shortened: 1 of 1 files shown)


sspanak:
> I noticed the app was not handling an Action ID of 6 (IME_ACTION_DONE).

@alexknop, this is not correct. `IME_ACTION_DONE` is supposed to be handled in `case IME_ACTION_ENTER` below. Please check find `textField.getAction()` just above the `switch..case` and check what it returns. If it returns something different than `TextField.IME_ACTION_ENTER` either there is something wrong with it and the fix has to be applied there (and not in `TraditionalT9.java`), or the application has set the `IME_FLAG_NO_ENTER_ACTION` which means "do not send ENTER".

alexknop:
It returns 6, which is why I added the extra case.
IME_ACTION_ENTER is equal to 256

sspanak:
> It returns 6, which is why I added the extra case.

Yes, that's the point, `textField.getAction()` should never return 6. If you look at its code, you would see, other actions such as "GO", "NEXT", "SEARCH" or "SEND" are allowed, but all other cases are converted to "ACTION_ENTER". So, if it returns "6", then something is wrong with that function.

Possibly, the text field could have a custom actionId "6", instead of imeOption "6", which would result in that function returning 6. If that is the case, we'll figure out how to proceed.

Either way, as I said, `TraditionalT9.java` class is working fine. The fix must be applied in the `TextField.java` file.

alexknop:
Yes, when getAction() is called, field.actionId is 6.
So when it hits this line:
`
if (field.actionId > 0) {
			return field.actionId; // custom action, defined by the connected app
		}
`
then it just returns 6 to TraditonalT9.Java

sspanak:
I see. Then, I would like to suggest a slightly different approach.

1. Let's revert the changes in `TraditionalT9.java`.
2. Then apply the same logic in `TextField.java` and replace `getAction()` with this:
```java
	public int getAction() {
		if (field == null) {
			return EditorInfo.IME_ACTION_NONE;
		}

		int action;
		if (field.actionId > 0) {
			// Custom action, defined by the connected app. Even though some apps tend to pass a standard
			// EditorInfo.IME_ACTION_XXX action using this property instead of "imeOptions".
			action = field.actionId;
		} else {
			action = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);
		}

		switch (action) {
			case EditorInfo.IME_ACTION_UNSPECIFIED:
			case EditorInfo.IME_ACTION_DONE:
				return IME_ACTION_ENTER;
			default:
				return action;
		}
	}
```

No application on my phone uses `actionId`, so I can't test this myself. Could you please try it out on your phone?

alexknop:
It works well. Done

sspanak (approved):

PR #310, opened 2023-07-20
added workaround to show the window for older Android devices < 9

alexknop:
Problem:
As previously asked, there did not seem to be a method to handle the keyboard's main view from not showing. I have tested this method for my Sonim (which runs Android 8.1 and could not take advantage of requestShowSelf) and it is working now.
Solution:
Call showWindow(), which makes a call to showWindowInner() with the boolean you pass. When you pass true, attributes such as mShowInputRequested and mWindowVisible become true, which help assist in functions like updateInputViewShown()

### DIFF
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -795,15 +795,16 @@ public class TraditionalT9 extends KeyPadHandler {
 	 * are invisible. This function forces the InputMethodManager to show our window.
 	 */
 	protected void forceShowWindowIfHidden() {
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
-				&& !mInputMode.isPassthrough()
-				&& !isInputViewShown()
-		) {
+		if (mInputMode.isPassthrough() || isInputViewShown()) {
+			return;
+		}
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
 			requestShowSelf(InputMethodManager.SHOW_IMPLICIT);
 		} else {
-			super.showWindow(true);
+			showWindow(true);
 		}
 	}
+
 
 	@Override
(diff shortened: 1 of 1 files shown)


src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -805,9 +805,11 @@ protected void forceShowWindowIfHidden() {
 		) {
 			requestShowSelf(InputMethodManager.SHOW_IMPLICIT);
 		}
+		else{
+			super.showWindow(true);
+		}
```
sspanak: Shouldn't this be:
```suggestion
		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
			requestShowSelf(InputMethodManager.SHOW_IMPLICIT);
		} else {
			super.showWindow(true);
		}
```
?

The Main View must definitely not be displayed in Mode Passthrough. That's the one used when dialling a number or other scenarios when the keyboard must remain invisible.

sspanak:
I am not sure what happened here, but your code went back to the original state. Perhaps, I my suggestion was a bit unclear, because Github didn't allow me to select the lines above. Anyway, what I meant is, let's do this:

```java
	protected void forceShowWindowIfHidden() {
		if (mInputMode.isPassthrough() || isInputViewShown()) {
			return;
		}

		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
			requestShowSelf(InputMethodManager.SHOW_IMPLICIT);
		} else {
			super.showWindow(true);
		}
	}
```

I suppose there is no need to reference `super` explicitly, because there is no `showWindow()` function here, so you may want to try removing it. Also, I am not sure if `isInputViewShown()` messes up things on your phone. If my suggestion doesn't work, please try without it and let me know of the results.

alexknop:
that worked as well, thank you for the suggestions. New commits are pushed

sspanak (approved):

PR #311, opened 2023-07-21
changed onBackspace to first shorten the digitSequence and then deter…

alexknop:
Problem:
Even when deleting the first letter in a word, getSuggestions() is being called because onBackspace was returning true. The concern here was that digitSequence and suggestions need to be clear before beginning to type another word.

Solution:
Shorten digitSequence by 1 immediately. This makes sure that when deleting the last letter, the digitSequence is "".
suggestions are already cleared by onNumber() in ModePredictive.java. It calls for super.reset(), which clears suggestions.
This solution will make the code not run an extra step of trying to load suggestions which do not exist.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -300,12 +300,7 @@ public class TraditionalT9 extends KeyPadHandler {
 			return performOKAction();
 		}
 
-		String word = suggestionBar.getCurrentSuggestion();
-
-		mInputMode.onAcceptSuggestion(word);
-		commitCurrentSuggestion();
-		autoCorrectSpace(word, true, KeyEvent.KEYCODE_ENTER);
-		resetKeyRepeat();
+		acceptCurrentSuggestion(KeyEvent.KEYCODE_ENTER);
 
 		return true;
 	}
@@ -505,10 +500,10 @@ public class TraditionalT9 extends KeyPadHandler {
 		cancelAutoAccept();
 
 		if (delay == 0) {
-			this.onOK();
+			this.acceptCurrentSuggestion();
 			return true;
 		} else if (delay > 0) {
-			autoAcceptHandler.postDelayed(this::autoAccept, delay);
+			autoAcceptHandler.postDelayed(this::acceptCurrentSuggestion, delay);
 		}
 
 		return false;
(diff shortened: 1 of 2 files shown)


sspanak:
Calling `getSuggestions()` is completely safe with an empty `digitSequence`. The database is not queried and no heavy operations are executed, so it shouldn't cause any performance issues. Actually, it is a necessary step, to clear the suggestions after deleting the last letter.

I don't quite get what the problem is. Have you been experiencing problems with backspace?

alexknop:
It's not causing any issues, but if you delete the first letter of a word, you shouldn't be loading suggestions because there are none.
As I said, it's not necessary "suggestions are already cleared by onNumber() in ModePredictive.java. It calls for super.reset(), which clears suggestions."

sspanak:
> if you delete the first letter of a word, you shouldn't be loading suggestions because there are none.

Yes, I do. I need to "load" an empty list, in order to erase any one-letter suggestions that were displayed until now. And I need to update the state of the InputMode.

> suggestions are already cleared by onNumber() in ModePredictive.java. It calls for super.reset(), which clears suggestions."

Sure, but what if I just want to erase the current word? What you did would actually cause `mInputMode.getSuggestions()` called from `TraditionalT9` to return suggestions after erasing the `digitSequence`. But this could cause a bug in some edge case scenario. `getSuggestions()` must always return the correct suggestions for the current state.

Anyway, don't worry about this, there is no real "loading". I know it isn't obvious, but running `ModePredictive.loadSuggestions()` would actually do nothing, but clear an `ArrayList` then pass it back to the UI. No database queries, no data processing, no nothing. Indeed, after the suggestion list on the screen has been cleared, even that is unnecessary, but from my testing, it takes so little CPU time, that it is equivalent to falling back to the other case:
```
commitCurrentSuggestion(false);
super.sendDownUpKeyEvents(KeyEvent.KEYCODE_DEL);
```

So, please, let's not fix, what isn't broken.

alexknop:
good point, it does become dependent on onNumber to clear the suggestions...
what if we called reset after the last backspace of a word? that seems like a logical function instead of trying to load suggestions.

sspanak:
I am not sure how to put it another way, but this PR is not going in. I understand your concern, but I assure you no actual loading or processing of data takes place, despite calling `ModePredictive.loadSuggestions()`, so there is nothing to worry about.

PR #316, opened 2023-07-22
remove fadingEdge because it is deprecated

alexknop:
Problem:
XML attribute "fadingEdge" is applied to RecyclerView, however it is just being ignored:
 
        <attr name="fadingEdge">

More info: https://developer.android.com/reference/android/R.attr.html#fadingEdge

Solution: 
Remove unused xml attribute


sspanak:
Interestingly, I don't get a deprecation warning in Android Studio. 

@alexknop, what version do you use? Mine is "Flamingo" 2022.2.1 Patch 2. Have you changed in settings to enable the warning?

sspanak (approved):

PR #318, opened 2023-07-26
removed the last of the legacy Backspace hack

alexknop:
Problem:

When testing on a consumer cellular link ii where the clear key also doubles as the back key:
image and it is meant to delete text when there is text, otherwise double as a back key for system navigation.
I was not able to back out of a text conversation because I would be on a textbox which calls for the keypad and isBackspaceHandled was always returning true here. It wasn't returning true because of onBackspace() but because the keyCode was 67 (KeyEvent.KEYCODE_DEL) and was returning true for the second part of the OR statement.

Solution:

Removed second part of conditional. onBackspace() already checks whether there's text or if it's in passthrough. Open to discussing the logic here, but it doesn't seem necessary to have this OR statement.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
+++ b/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
@@ -122,7 +122,7 @@ abstract class KeyPadHandler extends InputMethodService {
 		if (Key.isBackspace(settings, keyCode)) {
 			// When there is no more text, allow "Back" key to function normally, not to block navigation.
 			// All other keys have their default function disabled.
-			isBackspaceHandled = onBackspace() || keyCode != KeyEvent.KEYCODE_BACK;
+			isBackspaceHandled = onBackspace();
 			return isBackspaceHandled;
 		} else {
 			isBackspaceHandled = false;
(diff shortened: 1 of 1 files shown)


src/io/github/sspanak/tt9/ime/KeyPadHandler.java
```diff
@@ -122,7 +122,7 @@ public boolean onKeyDown(int keyCode, KeyEvent event) {
 		if (Key.isBackspace(settings, keyCode)) {
 			// When there is no more text, allow "Back" key to function normally, not to block navigation.
 			// All other keys have their default function disabled.
```
sspanak: This comment is no longer valid, so let's delete it, too.
```suggestion
```

sspanak:
A long time ago, all the logic for the combined del/back keys used to be in `onKeyDown()`. And with time, I slowly refactored the code and moved the logic to more appropriate places such as `onBackspace()` and the `Key` class.

For a moment I thought this would break Backspace on my phone, because it works the opposite way. It has a true BACK key (code=4), that is also used for deleting text, when there is text, but `Key.isBackspace()` and `onBackspace()` handle everything properly now, so it's finally safe to delete the part of the hack.

This is absolutely useless information, unless you were curious what's with this line of code. :smile: 

Let's delete it, but please, also delete the comment above, as I noted in the other comment.

PR #325, opened 2023-07-29
Correct OK action when the connected app requests a standard IME action.

sspanak:


sspanak:
@alexknop, I've found that #309 actually breaks sending messages in Telegram for me. I've reverted the part about the standard action, but kept the `actionId == EditorInfo.IME_ACTION_DONE` part, that fixed several problems on your phone.

Yet, could you please test this branch? I have no means of verifying if your messaging and contacts app still work fine...

alexknop:
It works for the selecting a contact from the drop-down, but it does not work to send a message in my messaging app anymore.
When I press OK to send a message, field.actionId = 0 and field.imeOptions = 1140850692
Then after hitting `int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);`
standardAction is = 1073741828 which I think is invalid for the switch statement. 

Why isn't it working in Telegram anymore?

sspanak:
> ... standardAction is = 1073741828 which I think is invalid for the switch statement.

This means the "IME_FLAG_NO_ENTER_ACTION" bit is on, so [as per the documentation](https://developer.android.com/reference/android/view/inputmethod/EditorInfo#IME_FLAG_NO_ENTER_ACTION), the IME should not perform an action when ENTER/OK is pressed. "Typically this is because the action has such a significant impact or is not recoverable enough that accidentally hitting it should be avoided, such as sending a message."

It's all good from what I understand. You should be using a context menu or some other button for sending messages. Or there is no alternative?

> Why isn't it working in Telegram anymore?

It doesn't work, because we have changed the default output of `getAction()`. Instead of returning "IME_ACTION_ENTER" when no action can be determined, we return the invalid action. In Telegram and FB Messenger (and maybe other apps), it results in toggling the MainView instead of typing a new line.

Sending a new line is also what others do (at least Google Gboard keyboard and OpenBoard), so it is definitely the way to go.

Sorry if this takes away some convenience, but TT9 is a universal keyboard and I'd rather keep it closer to the recommended standard. It is impossible for me to support every single exotic Android implementation by adding more and more hacks. And believe me, the hacks will be many, because the Android world is anything but standard. I will probably accept some, but not ones that contradict the standards completely.

Having all that said, it looks like OK is working as it should and there is nothing else to do here.

alexknop:
So I didn't think the value going into the switch statement looked accurate and I changed the code to `int standardAction = field.actionId & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);`
and it worked. I wonder if that was the intention the first time?

alexknop:
The available cases in the switch statement are values such as 3, 4, 7, etc. but standardAction is 1073741828 so it's never going to meet a case.

sspanak:
> The available cases in the switch statement are values such as 3, 4, 7, etc. but standardAction is 1073741828 so it's never going to meet a case.

Correct, it's because `imeOptions` is sort of a settings store, that holds 64 different settings. This large value means the `IME_FLAG_NO_ENTER_ACTION` bit is on. And when it is on it means the application has requested the keyboard not to perform any action when OK or ENTER is pressed. Again, there is more info in the Android developer documentation. Check the link I posted above.

PR #326, opened 2023-07-30
added OR operator for calculator app

alexknop:
Problem: When using the Calculator app, the status bar was showing and covering part of the UI. I felt it made more sense to use passthrough here, similar to the dialer. It also allows me to press * and insert a decimal point on the calculator app
image

Solution: in isDialer() function, also consider when the inputType is android.text.InputType.TYPE_CLASS_NUMBER

### DIFF
--- a/src/io/github/sspanak/tt9/ime/helpers/InputType.java
+++ b/src/io/github/sspanak/tt9/ime/helpers/InputType.java
@@ -51,7 +51,7 @@ public class InputType {
 		int inputType = field.inputType & android.text.InputType.TYPE_MASK_CLASS;
 
 		return
-			inputType == android.text.InputType.TYPE_CLASS_PHONE && field.packageName.equals("com.android.dialer") || inputType == android.text.InputType.TYPE_CLASS_NUMBER;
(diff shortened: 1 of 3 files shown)


src/io/github/sspanak/tt9/ime/helpers/InputType.java
```diff
@@ -51,7 +51,7 @@ public boolean isDialer() {
 		int inputType = field.inputType & android.text.InputType.TYPE_MASK_CLASS;
 
 		return
-			inputType == android.text.InputType.TYPE_CLASS_PHONE && field.packageName.equals("com.android.dialer");
+			inputType == android.text.InputType.TYPE_CLASS_PHONE && field.packageName.equals("com.android.dialer") || inputType == android.text.InputType.TYPE_CLASS_NUMBER;
```
sspanak: This is an overkill for just fixing the calculator. It will hide the UI in any numeric field, but this is unnecessary most of the time and more important, it will make the on-screen keyboard unusable when one actually needs to type numbers.

Isn't it possible to use the `packageName` and match it to your calculator app instead?

sspanak:
> Yes, I have added a placeholder for it to also be the calculator app.
I used contains because my package name was technically "com.android.calculator2" but I bet there are package names that are just "com.android.calculator"

Yeah, I think this is good enough. I can't test this, because my calculator has its own keyboard, but I believe the fix is fine.

I've also renamed `isDialer()` in your branch, because it no longer checks only for Dialer fields. I hope that's OK.

PR #335, opened 2023-08-02
Feature/add kiswahili

marvoh:
Kiswahili is a language spoken in East, Central and parts of Southern Africa with over 100 million speakers. 

This merge introduces the language to the application.


sspanak:
Thanks for suggesting another new language. I am happy to see such pull requests!

From technical perspective there is no problem with it, however I can see the word list contains very few words, only ~16000. This is really not enough even for simple everyday chat. I don't mind merging it this way, but I am sure it will not be useful at all.

So how about extending with one of these [Tanzanian](https://github.com/elastic/hunspell/blob/master/dicts/sw/sw_TZ.dic) or [Kenyan](https://github.com/elastic/hunspell/blob/master/dicts/sw/sw_KE.dic) wordlists? Or maybe even merge them all and get a universal Kiswahili dictionary? Could you please check them out and let me know if this is a good idea?

marvoh:
Thank you for your feedback . 

I think the mistake I did was to look into the existing languages some of which had frequency next to the words so I settled on looking for a corpus scrapped from news sources hence the number of words. 

I'll merge the two then update the pull-request.

sspanak:
Having the frequencies is nice and it will improve the typing experience. It's OK to keep it.

marvoh:
I have updated corpus so will re-open with new language files

PR #347, opened 2023-08-13
Allowing Space to insert words, cleanup on setSuggestions, remove using isThereText

alexknop:
Problems:

1. A user cannot type a partial word, see the suggestion as the first one, and press 0(Space) to insert it. It would only insert what they've partially typed. I do think it should act like OK and insert.
2. setSuggestions() had two functions, where 1 was just to call the other and include a needed parameter. I've cleaned up the duplicate code and am always including the parameter.
3. isThereText() was used alongside getTextBeforeCursor() which was rather redundant, since both check the TextField to see if there is existing text. isThereText is only used for one if statement in determineNextWordTextCase() and I changed it to instead utilize getTextBeforeCursor() so that isThereText can be unused.


sspanak:
> A user cannot type a partial word, see the suggestion as the first one, and press 0(Space) to insert it. It would only insert what they've partially typed. I do think it should act like OK and insert.

This is not a bug, it's a feature :smile:. If you want the entire word, press OK, if you want to cut it, press SPACE. I really don't get what's the problem here. On the other hand, with your changes, there is no option to type a part of a word and the OK action is duplicated.

Either way, I will not accept this particular change.

2. and 3. are fine, provided that they are tested and they work.

alexknop:
Why would a user type part of a word?

alexknop:
Makes sense, not all words are in there. I am going to remove the space to insert part and keep the rest because I believe that is worth looking into.

alexknop:
Okay, the Space key code is changed back, but please look at the other things which are just code cleanup

sspanak:
> Okay, the Space key code is changed back, but please look at the other things which are just code cleanup

Done and this time as was paying more attention. :slightly_smiling_face: 

Taking away `isThereText` indeed looks valuable. I am willing to test it a bit and if there are no issues within 24 hours, I will accept it. It's what I always do, even with my code.

Now, removing the `setSuggestions()` with argument - I don't aggree with that. Yes, from technical perspective it serves no purpose, but from human perspective it provides convenience. 

As I was writing the code, I found that it is annoying and sometimes unclear to use `setSuggestions(null, 0)`. "What does `null, 0` mean? I just want to clear the suggestions, why do I have to think what's the proper value for the second argument?" As I am thinking of something completely different and I just want to clear the suggestion list, I don't want to stop and ask myself such questions. And there are many other convenience functions in the code, not just this one. I've decided, if I am writing most of the code, I should feel most comfortable with it.

So let's keep both overloads of `setSuggestion()`, please.

sspanak:
It is good from my side too. I am currently stabilizing version 26.0, so no new changes are accepted, unless they fix something in #358. I'll test and merge this after that.

sspanak (approved):

GiorgosXou:
The  "space-input" feature is a really usefull and common one among old t9 layouts. It would be nice if there was an option for it

sspanak:
@GiorgosXou, merged pull requests are not the right place for making requests. Please, create a new issue whenever you want to do so.

To answer your question, accepting words and automatically adding space after them is possible with the OK key.

PR #349, opened 2023-08-14
fix handling of 1 key

alexknop:
Problem:

I was catching instances where adding the apostrophe when typing "I'm" was not working.
I stepped through this algorithm and found a flaw: handleSuggestions() has the potential to process a digitSequence of 416 depending on how fast you type. Thus producing Imim.
digitSequence of 4 -> I,G,H
digitSequence of 416 -> im,in,io (suggestions get cut to m,n,o in clearLastAcceptedWord() )
digitSequence of 416 again -> im,in,io

Screenshot_19700101-175512
Screenshot from 2023-08-14 13-59-59

I have corrected this process to properly process 1 regardless of how fast you type in order to return just punctuation marks on 1, and not i,i.i-, etc which was eventually being trimmed in clearLastAcceptedWord() to just punctuation marks.
I wrote the code to no longer use loadStatic() to load punctuation, since that was breaking up the digitSequence and thus handleSuggestions(), but instead punctuation is later loaded in generatePossibleCompletions()

Lastly, removed a duplicate commitCurrentSuggestion() that was just calling the other function.

### DIFF
--- a/src/io/github/sspanak/tt9/db/DictionaryDb.java
+++ b/src/io/github/sspanak/tt9/db/DictionaryDb.java
@@ -215,8 +215,22 @@ public class DictionaryDb {
 			language.getId(),
 			maximumWords,
 			sequence,
-			filter == null || filter.equals("") ? null : filter
+			filter.equals("") ? null : filter
 		));
+		// redacting a function that was added but no longer needed
+		// leaving in case threading needs to come back
+		// If user types fast and sequence has a 1 in it, such as #1##,
+		// try to look for exact matches of 1###, only if previous
+		// query returned 0 matches
+//		if (matches.size() == 0 && sequence.contains("1") && sequence.charAt(sequence.length()-1) != '1') {
+//			sequence = sequence.substring(sequence.indexOf("1"));
+//			matches = new WordList(getInstance().wordsDao().getMany(
+//				language.getId(),
+//				maximumWords,
+//				sequence,
+//				filter.equals("") ? null : filter
+//			));
+//		}
 
 		printDebug("loadWordsExact", "===== Exact Word Matches =====", sequence, matches, start);
 		return matches.toStringList();
@@ -258,12 +272,7 @@ public class DictionaryDb {
 	}
 
 
-	private static void sendWords(ConsumerCompat<ArrayList<String>> dataHandler, ArrayList<String> wordList) {
-		asyncHandler.post(() -> dataHandler.accept(wordList));
-	}
-
-
-	public static void getWords(ConsumerCompat<ArrayList<String>> dataHandler, Language language, String sequence, String filter, int minimumWords, int maximumWords) {
+	public static ArrayList<String> getWords(Language language, String sequence, String filter, int minimumWords, int maximumWords) {
 		final int minWords = Math.max(minimumWords, 0);
 		final int maxWords = Math.max(maximumWords, minimumWords);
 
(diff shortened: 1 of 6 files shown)


sspanak:
I like the idea of simplifying the logic around 1-key. And I agree the race condition should be fixed, even though it is not a problem on most devices. If you have experienced it, your phone must be lagging a lot.

Unfortunately, what you did breaks typing in French, Italian and possibly Ukrainian and Hebrew.

It is true in English 1-key can be only at the beginning of a combination (e.g. `'ve` = 183 or `'s` = 17), but in French and Italian it could be the other way around. For example, they have `d'` and `l'` that could be combined with any word, e.g. `l'argent`, `l'ordinateur`... 

And in Ukrainian and Hebrew, the apostrophes are sort of used like letters. They can be anywhere in a word and have a phonetic meaning, so it must be possible to type words with containing 1-key in the middle, without breaking them, for example: `пам'ять` or `אבוריג'יניות`.

Having this said, I can't allow merging, sorry. If you come up with an alternative that works for all languages, that would be OK, but the current version will do more harm than good.

sspanak (changes requested):
An alternative solution that works for all languages is necessary.

alexknop:
I am still poking around for a solution but one thing I think would help is while the Thread in getWords performs a lookup on given a digitSequence, if the current digitSequence gets updated (you press another key) before it is done looking up, that onDbWords() can handle that data separately. Sometimes by the time onDbWords() is ran, the digitSequence is different than what the results are for.
This would prevent handleSuggestions receiving a digitSequence of 416 (for I'm) with only words of I (4) in them to process.
I am learning about Threads, Runnables, etc....

alexknop:
This is what I mean, the sequence is 416 but the app is just now processing the outcome of 4 (dBwords of [h,g,I]:
image

alexknop:
Alrighty, so an update..
I have changed the ability to have threads for the various digitSequences. Some processed faster than others when I type too fast and therefore weren't returning in the correct order to load suggestions.
Putting these digitSequences in sequential order has helped me a ton.

I've also gone through various parts of the code and made other optimizations, please take a look and let me know what you think.

sspanak:
> I have changed the ability to have threads for the various digitSequences. Some processed faster than others when I type too fast and therefore weren't returning in the correct order to load suggestions.
Putting these digitSequences in sequential order has helped me a ton.

Yes, that's what I meant when I said there is a race condition. The queries for digit sequences of length 2 are much slower compared to all other queries. And when two asynchronous processes depend on each other, but they complete in the incorrect order, yielding an incorrect result, that's called a race condition. I thought you knew what I mean. :slightly_smiling_face:

So, I took my 10-year old tablet with 8Gb storage, tried how typing works and I must admit it is not so good. It's not only the words with apostrophes. Typing anything faster produces wrong results, say "668" for "not", actually results in "668" results flashing for a moment, but then the ones for "66" appear. So, instead of "not", I see "on".

This definitely needs fixing, but I don't think it is worth investing any more effort in RoomDB+SQLite. I am thinking of replacing switching to another database. SQLite is not the perfect choice for storing this kind data anyway. But even if it was, there are alternatives that work much, much faster (or so they promise...).

So your code does not really fix the problem with the database slowness. Running the queries on the UI thread actually makes your phone unresponsive until each query completes. This indeed allows the queries to run in the correct order, but it has other downsides. There is no guarantee any query will finish in a reasonable amount of time. And while it is running, the phone will not respond to anything, which may lead to user frustration. Now, Android will not allow any application to block the phone for very long and will eventually kill it. As you can imagine, people will start complaining it is not working at all, which is even worse.

All in all, running on the UI thread is not a good idea and I will not accept it. You could keep your code if you want to, I do not plan any big changes in the near future, so it would be easy to sync with `master` and stay up-to-date. But in the long term, I'd rather switch to another database or at least a different, faster DAO.

alexknop:
Makes sense, from what I was reading it is not a best-practice to run this on the main thread either. I was putting on a band-aid to try and make this return in order. 
I will close this PR, however I think my code cleanup changes can be merged, so I will create a new PR

PR #355, opened 2023-08-17
Tidying Up Code.

alexknop:
Problems:
1.  "filter == null || filter.equals("") ? null : filter" If filter was null, it would return null anyways. So removed that conditional
2. "commitCurrentSuggestion()", used once, just pointed to commitCurrentSuggestion(true) so I changed the 1 usage and removed this parameter-less function.
3. learned that the correct situation to add a space after a word is:
a. When "Automatic Space" is on in settings
b. User presses OK
The function to check whether to add a space was called in many other scenarios, such as when the user types in a certain digitSequence and then 0. We are already committing to the previous word in this scenario, so it is not needed.
We also were calling it in onText, and if a user types in * and it's not a hotkey, for example, then we are again already committing the previous word and inserting *. I don't see a need to add a space after the * here.
Lastly, we are calling it in handleSuggestions, but if we are inserting a symbol in our digitSequence, we don't want it to add a space as well at the time of pressing since we are again committing the previous word. Only if the user presses OK on the symbol (which is handled anyway).
4. Speaking of handling suggestions, we already check in shouldAcceptPreviousSuggestion() if the sequence has 1 and another digit besides 1, so when it comes to clearLastAcceptedWord(), some of our checks here are always going to pass: lastAcceptedWord wont be empty, suggestions wont be empty, and a suggestion's length will always be greater than the previous word's length.
5. You changed the algorithm for apostrophes to be able to remove apostrophe (English) words in the dictionary, so I see insertPunctuationCompletions() as an unnecessary function now.
6. This one could be debated, but I thought it would be better to just use 1 character for learning "previousChars". In using 2 characters for the use cases of typing " -" or " /", I thought it was inaccurate to consider them a part of a digitSequence. If I was typing "and / or", I would not expect to be able to type 164 and for there to be a space added in between the "/" and "or", but rather I would consider inserting a "/" between two sequences/words.

Hope I explained these things well but if not let's discuss.

### DIFF
--- a/src/io/github/sspanak/tt9/db/DictionaryDb.java
+++ b/src/io/github/sspanak/tt9/db/DictionaryDb.java
@@ -212,7 +212,7 @@ public class DictionaryDb {
 			language.getId(),
 			maximumWords,
 			sequence,
-			filter == null || filter.equals("") ? null : filter
+			filter.equals("") ? null : filter
 		));
 
 		printDebug("loadWordsExact", "===== Exact Word Matches =====", sequence, matches, start);
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -540,7 +540,7 @@ public class TraditionalT9 extends KeyPadHandler {
 		}
 
 		mInputMode.onAcceptSuggestion(word);
-		commitCurrentSuggestion();
+		commitCurrentSuggestion(true);
 		autoCorrectSpace(word, true, fromKey);
 		resetKeyRepeat();
 	}
@@ -558,10 +558,6 @@ public class TraditionalT9 extends KeyPadHandler {
 		return currentWord;
 	}
 
-
-	private void commitCurrentSuggestion() {
-		commitCurrentSuggestion(true);
-	}
 
 	private void commitCurrentSuggestion(boolean entireSuggestion) {
 		if (!isSuggestionViewHidden()) {
@@ -629,7 +625,7 @@ public class TraditionalT9 extends KeyPadHandler {
 
 
 	private String getComposingText(int maxLength) {
-		if (maxLength == 0 || suggestionBar.isEmpty()) {
+		if (suggestionBar.isEmpty()) {
 			return "";
 		}
 
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -283,7 +283,7 @@ public class TraditionalT9 extends KeyPadHandler {
 		// First pass, analyze the incoming key press and decide whether it could be the start of
 		// a new word.
 		if (mInputMode.shouldAcceptPreviousSuggestion(key)) {
-			autoCorrectSpace(acceptIncompleteSuggestion(), false, key);
+			acceptIncompleteSuggestion();
 		}
 
 		// Auto-adjust the text case before each word, if the InputMode supports it.
@@ -333,12 +333,11 @@ public class TraditionalT9 extends KeyPadHandler {
 		forceShowWindowIfHidden();
 
 		// accept the previously typed word (if any)
-		autoCorrectSpace(acceptIncompleteSuggestion(), false, -1);
+		acceptIncompleteSuggestion();
 
 		// "type" and accept the new word
 		mInputMode.onAcceptSuggestion(text);
 		textField.setText(text);
-		autoCorrectSpace(text, true, -1);
 
 		return true;
 	}
@@ -541,7 +540,7 @@ public class TraditionalT9 extends KeyPadHandler {
 
 		mInputMode.onAcceptSuggestion(word);
 		commitCurrentSuggestion(true);
-		autoCorrectSpace(word, true, fromKey);
+		autoCorrectSpace(word, fromKey);
 		resetKeyRepeat();
 	}
 
@@ -592,7 +591,6 @@ public class TraditionalT9 extends KeyPadHandler {
 			String lastComposingText = getComposingText(mInputMode.getSequenceLength() - 1);
 			commitCurrentSuggestion(false);
 			mInputMode.onAcceptSuggestion(lastComposingText, true);
-			autoCorrectSpace(lastComposingText, false, -1);
 			mInputMode.determineNextWordTextCase(textField.isThereText(), textField.getTextBeforeCursor());
 		}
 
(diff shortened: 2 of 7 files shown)


src/io/github/sspanak/tt9/db/DictionaryDb.java
```diff
@@ -212,7 +212,7 @@ private static ArrayList<String> loadWordsExact(Language language, String sequen
 			language.getId(),
 			maximumWords,
 			sequence,
-			filter == null || filter.equals("") ? null : filter
+			filter.equals("") ? null : filter
```
sspanak, line 215: This may cause `NullPointerException`. You can't call `.equals()` on NULL.

src/io/github/sspanak/tt9/ime/modes/helpers/Predictions.java
```diff
@@ -255,46 +255,6 @@ private ArrayList<String> generatePossibleCompletions(String baseWord) {
...
 
 
-	/**
-	 * insertPunctuationCompletions
-	 * When given: "you'", for example, this also generates all other 1-key alternatives, like:
-	 * "you.", "you?", "you!" and so on. The generated words will be inserted after the direct
-	 * database matches and before the fuzzy matches, as if they were direct matches with low frequency.
-	 * This is to preserve the sorting by length and frequency.
-	 */
-	private ArrayList<String> insertPunctuationCompletions(ArrayList<String> dbWords) {
```
sspanak, line 265: You seem to have misunderstood what this does. It is the key to fixing #187. The punctuation marks may only be removed when accepting words, not when displaying the suggestions.

In short there were complaints that when one starts typing `it's`, it is not possible to type `it,`, because the database used to contain only `it'(s)`. This method makes sure all possible variants of `it` + punctuation mark are available. And this is not the only case. French contractions such as: `j'`, `l'`, `qu'`, were also causing the same problem. In fact, typing in French used to be very difficult, due to the large number of contractions they use.

Please, do not remove this unless you have a better idea how to type contractions or words with apostrophes in French, Italian, Ukrainian, Hebrew and others I can't recall at the moment.

src/io/github/sspanak/tt9/ime/modes/helpers/AutoSpace.java
```diff
@@ -84,8 +85,6 @@ private boolean shouldAddAfterPunctuation(String previousChars, String nextChars
 				|| previousChar == '%'
 				|| previousChar == '»'
 				|| previousChar == '“'
-				|| previousChars.endsWith(" -")
-				|| previousChars.endsWith(" /")
```
sspanak, line 88: What is the reason for removing these? Even if you were not expecting the automatic space, you would just not type it manually when you see it. The reasoning is: given that you separated the first word and the slash with a space, there is a very high chance that you would want a space before the second word too, so let's offer it automatically.

Please, note that there will be an automatic space only after: "and /" and not after: "and/".

src/io/github/sspanak/tt9/ime/modes/helpers/AutoSpace.java
```diff
@@ -31,6 +31,10 @@ public AutoSpace setLastWord(String lastWord) {
 		return this;
 	}
 
+	public boolean getAutoSpace() {
+		return settings.getAutoSpace();
+	}
+
```
sspanak, line 37: You have access to `SettingsStore` here too. 

But, either way, what is the benefit of this change? Currently, the setting check does not malfunction or block any future changes, and it fits right in its place. It is just fine as it is. I just don't see any benefit in moving it anywhere else.

src/io/github/sspanak/tt9/ime/modes/ModePredictive.java
```diff
@@ -117,21 +117,14 @@ public void reset() {
...
-		if (
-			lastAcceptedWord.isEmpty()
-			|| suggestions.isEmpty()
-			|| !suggestions.get(0).toLowerCase(language.getLocale()).startsWith(lastAcceptedWord.toLowerCase(language.getLocale()))
-		) {
-			return;
-		}
 
 		int lastAcceptedWordLength = lastAcceptedWord.length();
-		digitSequence = digitSequence.length() > lastAcceptedWordLength ? digitSequence.substring(lastAcceptedWordLength) : "";
```
sspanak, line 129: Running `.substring()`, with a number larger than the current string length, causes "index out of range exception". The last accepted word could and probably is longer than the current digit sequence (well, until the user types enough letters to make it longer).

Don't think it can't go wrong. I can and it does sometimes, so I had to add an extra safety check. :slightly_smiling_face:

src/io/github/sspanak/tt9/ime/modes/ModePredictive.java
```diff
@@ -117,21 +117,14 @@ public void reset() {
 	 * or stops silently, when there is nothing to do.
 	 */
 	private void clearLastAcceptedWord() {
-		if (
```
sspanak, line 120: This prevents unnecessary `ArrayList` operations and saves some memory and CPU. Suggestion list is updated frequently, so savings are worth it.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -333,12 +333,11 @@ public boolean onText(String text, boolean validateOnly) {
 		forceShowWindowIfHidden();
 
 		// accept the previously typed word (if any)
-		autoCorrectSpace(acceptIncompleteSuggestion(), false, -1);
```
sspanak, line 336: `onText()` is intended to accept any text not just "*" and "#", so it may be necessary to add a space. If you enable the on-screen keyboard, you would see it has dedicated "?" and "!" keys. In the future, there may be even more use cases for `onText()`.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -283,7 +283,7 @@ protected boolean onNumber(int key, boolean hold, int repeat) {
 		// First pass, analyze the incoming key press and decide whether it could be the start of
 		// a new word.
 		if (mInputMode.shouldAcceptPreviousSuggestion(key)) {
-			autoCorrectSpace(acceptIncompleteSuggestion(), false, key);
```
sspanak, line 286: Auto-correcting the space here is necessary when one breaks a word with punctuation or goes from punctuation to a word.

You don't see this very often when typing in English, because the app often switches to some contraction, but for example, you could type: "Oh, are you ok?" without needing to press SPACE after "Oh,", it will be automatically added when you start typing "are". You could even start a new sentence after the question mark, without typing space. This is especially useful in languages without contractions, like Slavic languages, German, Swedish, Norwegian, etc... where you almost don't need to use the 0-key to type a space.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -559,10 +558,6 @@ private String acceptIncompleteSuggestion() {
 	}
 
 
-	private void commitCurrentSuggestion() {
```
sspanak, line 562: I guess it's OK to remove this one, if it is used in one place only, but it is really no big deal.

sspanak:
There are so many issues with this PR that I must reject it. The title says "tidying up the code", but it actually changes the application behaviour in many cases, sometimes causing bugs, sometimes neglecting scenarios which will become very difficult or even fail from user perspective, with the new code. 

Please, consider TT9 is used by many people on many different devices, in many different configurations, while you seem to be concentrated on optimizing everything for your language and phone. The code may not be perfect, but I've done many things for a reason.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
alexknop: Gotcha, I will say it doesn't work too much nowadays due to the addition of 'd, 'm, 've, 'll, etc to the dictionary. I was only really able to use that feature on the few keys that aren't programmed to have that.

sspanak:
Rejecting this due to so many issues described above.

PR #361, opened 2023-08-23
remove onstop

alexknop:
onStop is called during onStart (opening UI) and onFinishInput (closing UI)
it consists of three functions
1. onFinishTyping() - closes autoAccept handler
2. clearSuggestions() - clears existing suggestions
3. statusBar.selectText() - changes status bar to show input mode

I've found that by removing onStop, when you leave and come back to type, these 3 functions are called anyway.
onFinishTyping() is called by onFinishInputView() when you leave the keyboard.
clearSuggestions() main tasks are to setSuggestions to null, which is done during onStartInput when you open the keyboard back up and to set composing text to "". I've found this is not making a difference when you leave the keyboard anyway, as this existing suggestion gets committed anyway.
statusBar.selectText is called when you open the keyboard again.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
+++ b/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
@@ -93,17 +93,6 @@ abstract class KeyPadHandler extends InputMethodService {
 		onFinishTyping();
 	}
 
-	/**
-	 * This is called when the user is done editing a field. We can use this to
-	 * reset our state.
-	 */
-	@Override
-	public void onFinishInput() {
-		super.onFinishInput();
-		// Logger.d("onFinishInput", "When is this called?");
-		onStop();
-	}
-
 
 	/**
 	 * Use this to monitor key events being delivered to the application. We get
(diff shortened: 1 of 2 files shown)


src/io/github/sspanak/tt9/ime/KeyPadHandler.java
```diff
@@ -93,17 +93,6 @@ public void onFinishInputView(boolean finishingInput) {
 		onFinishTyping();
 	}
 
-	/**
```
sspanak, line 96: Please, don't do this. Applications are not very consistent in calling `onFinishInput()` and `onFinishInputView()`. They call the functions both when going to background mode and when the user switches to another input field in the same app. The thing is, in each scenario, applications could call either `onFinishInput()`, or `onFinishInputView()`, or both, despite the fact the documentation clearly states what they are for and when to call them.

This is the most stable solution I've found. It has been working for months without complaints, so let's not change anything.

alexknop:
That is a good point and I understand your concern with wanting to cover all bases when you don't know how an application is going to call the InputMethod Service. Although this is going back to some of the fundamentals of the app, I do think it's worth a revisit and seeing where things could be optimized. I am going to open a separate PR that keeps onStop but does tweak some other things that I think are worth considering. I just want to avoid any unnecessary processing. Thanks

PR #362, opened 2023-08-23
fixed duplicate tasks

alexknop:
Found some duplicate tasks
1. onKeyAddWord: calling textField.finishComposingText() and clearSuggestions(), but in clearSuggestions(), textField.finishComposingText() is called too
2. onKeyNextLanguage: commitCurrentSuggestion() is called which also runs the same tasks as clearSuggestions, and then clearSuggestions() is being called after that. I am sure there is some re-factoring of this logic that can be done, but for now just focusing on removing duplicate tasks

### DIFF
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -354,7 +354,6 @@ public class TraditionalT9 extends KeyPadHandler {
 		}
 
 		cancelAutoAccept();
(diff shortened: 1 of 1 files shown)


sspanak:
>  I am sure there is some re-factoring of this logic that can be done, but for now just focusing on removing duplicate tasks

That's fine. This is the result of refactoring the class many times. Missing a line or two is inevitable.

sspanak (approved):

PR #365, opened 2023-08-28
added new flag and added text check

alexknop:
So on my screen such as this, I was trying to press OK to select an app.
Screenshot_20230828-001103
Because of the included search bar, it made things tricky. My imeOptions were 134217733, which resulted in it getting filtered to IME_ACTION_NEXT, but IME_ACTION_NEXT was not selecting anything.
By including this extra IME_FLAG, it is getting filtered to its true value and then defaulting to IME_ACTION_ENTER.
However, when it got to oldText and newText, both old and new texts were "" and they were equal. This ran KEYCODE_DPAD_CENTER which was selecting the app, but then running KEYCODE_ENTER which was also valid to select something again on the next screen. My phone was acting as if I double clicked the OK button.

Assuming this logic here is used to send messages, I added a check to make sure oldText is not "". If it is, there isn't anything in a text message box, so don't perform the system ENTER action.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/TraditionalT9.java
+++ b/src/io/github/sspanak/tt9/ime/TraditionalT9.java
@@ -732,19 +732,20 @@ public class TraditionalT9 extends KeyPadHandler {
 				String oldText = textField.getTextBeforeCursor() + textField.getTextAfterCursor();
 
 				sendDownUpKeyEvents(KeyEvent.KEYCODE_DPAD_CENTER);
-
-				try {
-					// In Android there is no strictly defined confirmation key, hence DPAD_CENTER may have done nothing.
-					// If so, send an alternative key code as a final resort.
-					Thread.sleep(80);
-					String newText = textField.getTextBeforeCursor() + textField.getTextAfterCursor();
-					if (newText.equals(oldText)) {
-						sendDownUpKeyEvents(KeyEvent.KEYCODE_ENTER);
+				if (!oldText.isEmpty()) {
+					try {
+						// In Android there is no strictly defined confirmation key, hence DPAD_CENTER may have done nothing.
+						// If so, send an alternative key code as a final resort.
+						Thread.sleep(80);
+						String newText = textField.getTextBeforeCursor() + textField.getTextAfterCursor();
+						if (newText.equals(oldText)) {
+							sendDownUpKeyEvents(KeyEvent.KEYCODE_ENTER);
+						}
+					} catch (InterruptedException e) {
+						// This thread got interrupted. Assume it's because the connected application has taken an action
+						// after receiving DPAD_CENTER, so we don't need to do anything else.
+						return true;
 					}
-				} catch (InterruptedException e) {
-					// This thread got interrupted. Assume it's because the connected application has taken an action
-					// after receiving DPAD_CENTER, so we don't need to do anything else.
-					return true;
 				}
 
 				return true;
--- a/src/io/github/sspanak/tt9/ime/helpers/TextField.java
+++ b/src/io/github/sspanak/tt9/ime/helpers/TextField.java
@@ -335,7 +335,7 @@ public class TextField {
 			return field.actionId;
 		}
 
-		int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);
+		int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION | EditorInfo.IME_FLAG_NAVIGATE_NEXT);
 		switch (standardAction) {
 			case EditorInfo.IME_ACTION_GO:
 			case EditorInfo.IME_ACTION_NEXT:


sspanak:
> Assuming this logic here is used to send messages, I added a check to make sure oldText is not "". If it is, there isn't anything in a text message box, so don't perform the system ENTER action.

This is a correct assumption and I have indeed missed this case. I have moved and extended the OK action code to a new class, so your changes can not be merged directly, but I have included the new `if (!oldText.isEmpty())` in the new code and I believe the double action problem is fixed there too.

As for your problem with selecting an application, selecting something with OK sounds a bit strange to me. Is this how your built-in keyboard works? Looking at the screenshot, I would expect to use the DOWN key to get out of the search field and select the desired application, then use OK to confirm the selection. Doesn't that work for you? Or maybe the search field requires a DPAD_CENTER event to switch focus to the list below?

alexknop:
OK is used to confirm my selection, I still use the D-Pad to move up/down to select. I was reading a bit into IME_ACTION_NEXT and it seems to be an action where you can submit what you type and then move to another text box beneath, such as being in a contact page and entering First name, pressing OK, and there being a Last name box underneath that you can jump to next. I don't see a second box, and it's likely the imeOptions for this app are not programmed correctly.
You can move down and select any app and still there's a cursor on the top search box (seen in the screenshot), meaning you can type at anytime. Hope this made sense

sspanak:
> I was reading a bit into IME_ACTION_NEXT and it seems to be an action where you can submit what you type and then move to another text box beneath, such as being in a contact page and entering First name, pressing OK, and there being a Last name box underneath that you can jump to next. I don't see a second box, and it's likely the imeOptions for this app are not programmed correctly.

Probably, it is programmed incorrectly. There are many such apps in the Android world. Also, that's the reason `performOkAction()` is overcomplicated - I tried to handle as much messaging apps as possible.

> OK is used to confirm my selection, I still use the D-Pad to move up/down to select. 
...
You can move down and select any app and still there's a cursor on the top search box (seen in the screenshot), meaning you can type at anytime. Hope this made sense

So the app is completely usable and TT9 is working as I would expect it to. I suggest we keep things the way they are. I don't like introducing unnecessary changes to the code.

Your other fix is in [this commit](https://github.com/sspanak/tt9/pull/358/commits/898750263dd2010a15519ded227848da24a38aae). I am going to merge it very soon.

sspanak:
> My imeOptions were 134217733, which resulted in it getting filtered to IME_ACTION_NEXT, but IME_ACTION_NEXT was not selecting anything.
By including this extra IME_FLAG, it is getting filtered to its true value and then defaulting to IME_ACTION_ENTER.

This is not the true value. `imeOptions` 134217733 = `5 | 134217728` = `IME_ACTION_NEXT | IME_FLAG_NAVIGATE_NEXT`. So, `IME_ACTION_NEXT` is the actual true value.

You are mixing an IME option with IME action and getting invalid value which accidentally results in `IME_ACTION_ENTER`, because the `switch` statement below converts any unrecognized action to that.

I got curious about this flag and did a short research. According [to this discussion at Mozilla project](https://bugzilla.mozilla.org/show_bug.cgi?id=720475), `IME_FLAG_NAVIGATE_NEXT` and `IME_FLAG_NAVIGATE_PREVIOUS` are some obscure features that were never widely used. However, the authors of your app have decided to use the flag in a non-standard way. I am saying it is non-standard, because there is no next field to navigate to, but probably it signals something to the native keyboard. 

Anyway, for such strange use cases, I have created the [AppHacks class](https://github.com/sspanak/tt9/blob/master/src/io/github/sspanak/tt9/ime/helpers/AppHacks.java). Instead of making a change that would affect all applications, I would suggest adding a new case in `AppHacks.onEnter()`, for example:
```java
	public boolean onEnter() {
		if (isTermux()) {
			sendDownUpKeyEvents(KeyEvent.KEYCODE_ENTER);
			return true;
		} else if (isMessenger() || isMessengerLite()) {
			return onEnterFbMessenger();
		} else if (isTheSonimApplicationsList() { // name this as appropriate
			sendDownUpKeyEvents(KeyEvent.KEYCODE_ENTER); // or KEYCODE_DPAD_CENTER? If it is KEYCODE_ENTER, you could potentially merge it with the first case for Termux.
			return true;
		}

		return onEnterDefault();
	}
```

And this would be the function for recognizing the particular text field:
```java
private boolean isisTheSonimApplicationsList() {
	// validate as appropriate, this is just an example
	return isAppField("sonim.app.package.name", EditorInfo.TYPE_CLASS_TEXT) && editorInfo.imeOptions == (EditorInfo.IME_ACTION_NEXT | EditorInfo.IME_FLAG_NAVIGATE_NEXT);
}
```

I believe this would do the trick and it will not affect any other application.

src/io/github/sspanak/tt9/ime/TraditionalT9.java
```diff
@@ -732,19 +732,20 @@ private boolean performOKAction() {
...
 				sendDownUpKeyEvents(KeyEvent.KEYCODE_DPAD_CENTER);
-
-				try {
-					// In Android there is no strictly defined confirmation key, hence DPAD_CENTER may have done nothing.
-					// If so, send an alternative key code as a final resort.
-					Thread.sleep(80);
-					String newText = textField.getTextBeforeCursor() + textField.getTextAfterCursor();
-					if (newText.equals(oldText)) {
-						sendDownUpKeyEvents(KeyEvent.KEYCODE_ENTER);
+				if (!oldText.isEmpty()) {
```
sspanak, line 735: Please revert the changes in this file. As I said, I have already applied them in AppHacks.java

src/io/github/sspanak/tt9/ime/helpers/TextField.java
```diff
@@ -335,7 +335,7 @@ public int getAction() {
 			return field.actionId;
 		}
 
-		int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);
+		int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION | EditorInfo.IME_FLAG_NAVIGATE_NEXT);
```
sspanak, line 338: This is not the appropriate place for changing the action. Please follow check [my previous comment](https://github.com/sspanak/tt9/pull/365#issuecomment-1698595085) for more details

sspanak (changes requested):

alexknop:
This could work and I will test, it is a safer workaround than affecting all applications. Thanks for drawing up this solution

alexknop:
This did not work because of this line
`int standardAction = field.imeOptions & (EditorInfo.IME_MASK_ACTION | EditorInfo.IME_FLAG_NO_ENTER_ACTION);"`
because my imeOptions are 134217733, it is running through this line and getting reduced to 5 because there is not a `| EditorInfo.IME_FLAG_NAVIGATE_NEXT `option to check for that bit. 
I want it to default to Enter, not match with Next because the app is not authored correctly. Therefore when it returns to this line
`return action == TextField.IME_ACTION_ENTER ? appHacks.onEnter() : textField.performAction(action);`
my action is 5 and so appHacks is not called.
By adding `| EditorInfo.IME_FLAG_NAVIGATE_NEXT `, it was retaining its value of 134217733 and not matching with Next, but defaulting to Enter.

sspanak:
OK, this should be very easy to fix. Let's leave the pull request open, so that I don't forget about it. I intend to include this in the next version.

sspanak:
@alexknop, could you please provide the package name of the app and the input class of the search field, so that I can crate the new app hack?

alexknop:
@sspanak first of all, welcome back. Hope all is well.
package name is com.android.setttings and inputType has a value of 524289 which appears to be a combination of TYPE_TEXT_FLAG_NO_SUGGESTIONS & TYPE_CLASS_TEXT

alexknop:
The only issue I might see with it being in just AppHacks is that my action returns 5, which doesn't match the if statement of whether it is equal to 256 and if so, to run AppHacks:
image

sspanak:
> The only issue I might see with it being in just AppHacks is that my action returns 5, which doesn't match the if statement of whether it is equal to 256 and if so, to run AppHacks

So it's time for the AppHacks class to handle one more event. :slightly_smiling_face: Don't worry about it.

> package name is com.android.setttings

Is it really "setttings", with triple "T"? If so, that's great, because we can consider it unique.

If not, it's too generic and will probably match the Settings app on many devices, which is not desirable. We have to identify it in another way. Could you also copy the `===> Start Up` log from `KeyPadHandler.onStartInput()`? It should be the very first message in Logcat after the text cursor appears. Also could you please tell me what this prints:
```java
Logger.d("deviceId", "model: " + Build.MODEL + "; product: " + Build.PRODUCT + "; device: " + Build.DEVICE + "; manufacturer: " + Build.MANUFACTURER + "; brand: " + Build.BRAND + "; host: " + Build.HOST);
```
?

alexknop:
Ha, it is not with 3 T's, that was just a typo.
Here you go,

```2023-12-17 16:18:35.540  4904-4904  tt9/KeyPadHandler       io.github.sspanak.tt9                D  ===> Start Up; packageName: com.android.settings inputType: 524289 fieldId: 16908298 fieldName: null privateImeOptions: null imeOptions: 6 extras: null```


```model: XP5800; product: XP5812; device: XP5812; manufacturer: Sonimtech; brand: SONIM; host: build```

sspanak:
Excellent! Thanks!

sspanak:
@alexknop, I have approached the problem in an entirely different way. If you are still using the Sonim phone, could you please try out [v29.15](https://github.com/sspanak/tt9/releases/download/v1.0-inital/tt9-v29.15.apk)? Or, if you prefer to build the APK on your own, use the `redo-ok` branch.

Any feedback will be much appreciated!

Issue #366, opened 2023-08-28
Change the database for better dictionary performance


sspanak:
On lowend devices with slow storage (8 Gb) and old CPUs, loading the suggestions while typing could be quite slow. In particular, loading the fuzzy suggestions for two-letter words, could take 200-280 ms, on a Samsung SM-T230. This is also true for Sonim XP3 and other devices with similar specifications. 

A Slow database makes typing painfully slow as well and it could even lead to incorrect results. For example, if one tries to type a three-letter word, the two-letter suggestions may load too slow and replace the three-letter ones, leaving the user with invalid choices for the key presses. Another example is trying to type "I'm" in English, which could result in "Imim", because the three-letter suggestions depend on the two-letter ones, but they load in the wrong order.

Another problem is dictionary loading speed, which could take minutes in some cases.

The current storage mechanism, SQLite + RoomDb, can not provide much faster loading speed, besides, it is not the most optimal choice for storing the words. So let's switch to a different storage to optmize the performance.

* ~~Remove RoomDb~~
* ~~Adopt [Objectbox](https://objectbox.io/). According to random benchmarks, it offers at least \~4x faster reading and \~10x faster insert performance.~~
* ~~Ensure `DictionaryDb` interface remains the same. The rest of the application does not need to know about the database change.~~
* ~~Add a flag for distinguishing custom added words.~~
* ~~Consider separating the words for each language in a separate document~~ - No need to do it at the moment. Queries are reasonably fast and it will require separate models for each document, which will make adding new languages more difficult.
* ~~Ensure "Delete All" does not cause a crash while loading is in progress.~~
* ~~Figure out a migration path for the custom added words. The SQLite database would still be there, so it would be possible to recover them. The migration might take some time, so it would be better if the users decide whether to do it or not, instead of blocking their phones for no apparent reason.~~ - Will be handled separately, in [here](https://github.com/sspanak/tt9/tree/sql-words-recovery)

If Objectbox happens to be too difficult to use (highly unlikely), keep SQLite, but use [GreenDao](https://greenrobot.org/greendao/) ORM instead of RoomDb. Random benchmarks show a potential 1.5x speed improvement.

Doing this will make #53 obsolete.

sspanak:
It looks like Objectbox is quite fast, but it takes way too much storage space. 

Here are the statistics:
[...]

PR #379, opened 2023-10-09
czech wordlist

lukascalda:
Hi, 
I would like to contribute with czech wordlist.


lukascalda:
Job done, czech t9 keyboard is working already on my phone.

sspanak:
Thanks for your contribution!

I can see you have uploaded two word lists, a smaller one and a larger one and you are making use of the smaller one. What is the reason for this? Looking at the smaller one, it contains 96k words, which is very little for functional everyday usage. Should they be merged?

lukascalda:
Hi, no merging. The .csv one, the one yml refers to, is the right one. I had difficulties with the worldlist and I was trying different approaches. Those 96k words are ok, I am using the keyboard as is everyday since I made my build. It works amazingly good. Some words sometimes need to be added to dictionary, still it is perfectly functional as is by now.
I did not look yet if it possible to get user wordlist from the TT9 app. If it is, I will merge that one later with this already functional csv.
Cheers

sspanak:
> Hi, no merging. The .csv one, the one yml refers to, is the right one. I had difficulties with the worldlist and I was trying different approaches. Those 96k words are ok, I am using the keyboard as is everyday since I made my build. It works amazingly good. Some words sometimes need to be added to dictionary, still it is perfectly functional as is by now.

That's my point. It will work fine with the words in the dictionary, but you will need to add new ones all the time. That's what I thought it may be better using the larger list. 

> I did not look yet if it possible to get user wordlist from the TT9 app. If it is, I will merge that one later with this already functional csv.

I am fine with merging it as it is and upgrading later. Unfortunately, exporting within the app is not possible at the moment. Filtering only the custom words will take significant amount of time and processing power, so I decided not to implement it. This may change after #366, which will hopefully boost the database performance.

However there may be a way. I suppose you are now using a debug variant of TT9, since you have packaged Czech yourself. Provided that you are also using Android Studio, you could start the app while your phone is connected to your computer, then you could export the `words` table from the database using Android Studio. After that, you could just send me the entire table export, even it contains words from other languages. I'll filter the new Czech words manually. I have handy scripts for the job.

PR #384, opened 2023-10-28
Add Lithuanian language and translation

mister-nemo:


sspanak:
Thanks for your contribution!

sspanak:
The dictionary contains invalid words an numbers. They need to be cleaned up before merging.

CptOctober:
Hey, Lithuanian here, how would one go about downloading this to use on their phone, as the main download doesn't have it yet. Even if it has any issues.

CptOctober:
Hey is there any way to download this onto my phone even though it might be a little unfinished? Like compile it into an apk or something?

sspanak:
> Hey is there any way to download this onto my phone even though it might be a little unfinished? Like compile it into an apk or something?

This branch is based on v26.0, which is quite slow and contains many bugs which are already fixed. If you are OK with that, just checkout mister-nemo's branch and build the APK.

However, I would strongly recommend for you to wait for the official release. I haven't reviewed thoroughly the word list and I can't guarantee the quality of the suggestions.

sspanak:
> Yea I was more asking of how the hell do I build an APK xd

If you want to build your own APK, check out [the development guide](https://github.com/sspanak/tt9/blob/master/CONTRIBUTING.md#building). If you don't want to use Android Studio, you may go away by just installing Java and building it with `./gradlew assembleDebug`. 

But as I said, I don't recommend running such an APK, because it will be way slower than the official versions, and you won't be able to upgrade later without uninstalling.

> Also when could I be expecting a Lithuanian addition to the app?
Also, also, how could I contribute in making this thing go faster?

I am considering of including it in v29.0, which means it will be released approximately within the next 3 weeks to 2 months. There is not a company project, so there is no strict plan or schedule. I can't go any faster than this. Sorry!

CptOctober:
Thanks for the reply, I will be eagerly waiting and following :)
Might still try to compile my own.

PR #401, opened 2023-12-09
Summer bugs

### LINKED ISSUE #385: Occasional crash
I'm experiencing occasional crashes!
But it does not happen very often! They are actually somewhat rare.

They tend to happen when I focus some field. When the crash happens it usually happens 3 or 4 times one after the other for everytime I try to focus the field!!!

(sorry if I could not find a reason or a way to replicate yet...)

### LINKED ISSUE #389: No language and mode indication on start in some cases.
When opening any existing text file, note, telegram/watsapp chat, there is no information about which input language or mode is selected. The tray is not displayed before you press any button or touch the screen. And since there is no indication in the status bar anymore, you will not be able to determine which mode you will be typing in before you start typing. Sometimes it's annoying.

sspanak:
Closes #367: the MainView is visible in dropdowns or other non-text inputs

Fixed logging settings not being applied at startup sometimes

Closes #364: removed isActive

Fixes #385, Fixes #389: fixed a startup crash in onEvaluateInputViewShown()

### DIFF
--- a/res/values-bg/strings.xml
+++ b/res/values-bg/strings.xml
@@ -29,7 +29,7 @@
 	<string name="pref_category_appearance">Облик</string>
 	<string name="pref_category_function_keys">Бутони за бърз достъп</string>
 	<string name="pref_hack_fb_messenger">Изпращай съобщения с ОК в Messenger</string>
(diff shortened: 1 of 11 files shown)


sspanak:
Typing in Snapchat is broken

sspanak:
Also merge #347 into this.

PR #412, opened 2024-01-09
Arabic Language Pull Request

Arabi12321:
I created YML, csv dictionary, and txt wordlistreadme files for arabic language support on TT9, Please test and let me know if this would be able to be pulled.

IMPORTANT: THIS LANGUAGE READS RIGHT TO LEFT, thorough testing may be needed due to difference in right vs left hand justification.


languages/definitions/Arabic.yml
```diff
@@ -0,0 +1,16 @@
+---
+locale: ar-JO
+dictionaryFile: ar-utf8.csv
+abcString: ابت
+hasUpperCase: no
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION,؟,ى,ئ] # 1
```
sspanak: Letters on 1-key are not supported and may cause strange side effects while typing. It may be possible to make it work, since English, French, Hebrew and Ukrainian sort of use apostrophes and dashes as letters, meaning there are words with 1-key in their dictionaries, but there is no 100% guarantee that it will work properly.

Also, I don't see them included in standard hardware layouts, for example, [this Nokia](https://www.alamy.com/arabic-keyboard-on-cell-phone-in-hand-image343639432.html).

I suppose, they don't have a standard position? Would it be possible to move them to another key?

Arabi12321:
CSV file changed from spaces to tabs for frequency count

docs/dictionaries/arWordlistReadme.txt
```diff
@@ -0,0 +1,5 @@
+Source: https://github.com/a3f/arabic-wordlists
```
sspanak: Wikipedia is not really a good word source if used on its own. It may include a lot of technical terms, but it may lack commonly used everyday words.

Nevertheless, This is not a problem. There are many other free sources, like [this one](https://github.com/jamalamch/Arabic-Words/blob/master/AllListWordArabic.txt). The author claims to have gathered all possible words, and it is indeed an extensive list. I can include them before merging.

sspanak (commented):
> IMPORTANT: THIS LANGUAGE READS RIGHT TO LEFT

Right, the way you have structured the CSV confuses the validator. Make sure you put the words first, then the frequency. The words must be written from right to left, though, as one would normally write them.

> thorough testing may be needed due to difference in right vs left hand justification.

I am not sure what justification are you referring to. The only place Arabic text would be displayed in TT9 is the labels on the virtual keypad (if it is enabled). But they are centered, so it shouldn't matter. Or do you mean something else?

Finally, I can't read Arabic, so I won't be able to check for invalid or misspelled words. The word list will be used as it is. However, you are welcome to submit corrections later.

languages/definitions/Arabic.yml
sspanak: Arabic question mark must be included in `TextTools` for proper sentence end detection. Also, `AutoSpace.shouldAddAfterPunctuation()` for automatic space to work correctly.

Using the English punctuation chars and extending it with Arabic ones is not a good idea. Create a new punctuation list for Arabic only. Include the inverted comma `،` and semicolon `؛` there. 

[This](https://kaleela.com/en/punctuation-marks-in-arabic-and-how-theyre-used) looks like a good punctuation guide.

sspanak (changes requested):

languages/definitions/Arabic.yml
ghost: i added the arabic question mark to TextTools and AutoSpace (shouldAddAfterPunctuation and shouldDeletePrecedingSpace), and i have also created the punctuation list.

docs/dictionaries/arWordlistReadme.txt
ghost: I updated and briefly looked over the arabic word list you have recommended and it looks good, i left it in the format that it was downloaded and just renamed the txt file and updated the yml file to point to it, i have also updated the arWordlistReadme to reflect the new list. I edited the dictionary file to remove duplicates and standalone letters as well.

languages/definitions/Arabic.yml
ghost: i actually have an old samsung candy bar dumbphone from the mid 2000s in arabic and i have referenced it for the letters on the 3 key, the physical keyboard is identical to the picture you have linked but in actuality in the software the 3 key has a massive amount of letters linked to it, mostly different forms of alif (the first letter in the alphabet). the arabic keyboard is mapped from right to left, so physically the order of the alphabet on the keyboard maps to, in order, 3 2 6 5 4 9 8 7, the yml file has been fully mapped correctly to the standard layout. I have the phone so please ask any additional questions about how the original arabic keyboard works on these old arab T9 phones.

ghost: also I dont see the options for this in any of the files but even though the physical arabic keypad shows 1234567890, it types out as ١٢٣٤٥٦٧٨٩٠ (in that order, in arabic the numbers read from left to right just like english, even though the rest of the language is read right to left)

docs/dictionaries/arWordlistReadme.txt
ghost: I marked this conversation as resolved for now because I dont see any more work left to do for the dictionary files, please feel free to unresolve it if something comes up.

languages/definitions/Arabic.yml
ghost: the old candy bar phone i have has the punctuation's set as such: 
key 1: . (full stop), ؟ (Question mark), ، (Comma), ! (Exclamation), ؛ (Semicolon), " (Quote), @ (At), : (Colon), ١ (One)
key 0: + = < > € £ $ % & ¥ ¤ ٠
(# is mapped to space and * changes between language and numbers, holding * turns on T9 (predictive) and holding # gives a full menu with many new punctuation marks (i can provide images if needed))
all the keys will display their number as the last option, the ٠ for the 0 key is the arabic zero, not a dot.

ghost: also all the keys when long held will print their number without switching the language to number mode, just like the current state in TT9, sorry for the extensive messages, i just want to document the original arabic t9 keyboard layout as i have not found any reliable information online, the best source is the original i guess, if you want/need pictures/video i would be more than happy to provide you with that as well

ghost: I think i found all the files to change the punctuation in and have made the according changes.

Arabi12321 (commented):
I have completed all of the requested tasks so far, I think the only thing that remains is to add the arabic numbers and to test

Arabi12321:
Synced fork with upsteam

Arabi12321:
> > IMPORTANT: THIS LANGUAGE READS RIGHT TO LEFT
> 
> Right, the way you have structured the CSV confuses the validator. Make sure you put the words first, then the frequency. The words must be written from right to left, though, as one would normally write them.
> 
> > thorough testing may be needed due to difference in right vs left hand justification.
> 
> I am not sure what justification are you referring to. The only place Arabic text would be displayed in TT9 is the labels on the virtual keypad (if it is enabled). But they are centered, so it shouldn't matter. Or do you mean something else?
> 
> Finally, I can't read Arabic, so I won't be able to check for invalid or misspelled words. The word list will be used as it is. However, you are welcome to submit corrections later.

The reason i mentioned justification is that the TT9 software in abc mode shows the letter selection left to right in english when keys are being selected and was just wondering if it would automatically switch to right to left selections in arabic mode.

Arabi12321:
I dont see the options for this in any of the files but even though the physical arabic keypad shows 1234567890, it types out as ١٢٣٤٥٦٧٨٩٠ (in that order, in arabic the numbers read from left to right just like english, even though the rest of the language is read right to left). the java.util.Locale doesn't include any Arabic locales by default but I wonder if the system and/or input locale was changed to arabic that the numbers would automatically switch to the arabic versions.

Arabi12321:
Roboto or Roboto Lt don't support arabic, currently the dictionary file is in courier new i believe but would noto sans arabic be better? currently the dictionary loader is not loading the dictionary, throwing something like "��!'" on line 1 is not a valid word. The justification of the language gets corrected when the system locale is set to arabic, but the only issue is that when selecting in that mode, the right arrow will go left, and the left arrow will go right. Also for some reason the abcString is not doing anything and the mode select just displays the keys on the 2 key which is not correct.

Arabi12321:
I tried converting the dictionary file to utf-8 using libiconv but it instead went to UTF-8 with BOM, but this is an improvement as now, even though the dictionary loader on the phone is still not loading the dictionary, it actually displays the arabic letters on the fail on line one instead of ��!'. I think if the file were to be converted to pure UTF-8, without BOM, the dictionary would load and the only outstanding issue left would be the abcString not displaying.

Arabi12321:
I converted the dictionary file into pure UTF-8 and now the dictionary loads and predictive typing works!!!!! So that issue is completely resolved the only major issue that i would like some feedback on is about the abcString problem. Other than that, some minor things that can be added later to the Arabic keyboard is the arabic numbers and to fix when the phone locale is set to arabic, the letter/word options in TT9 are mapped wrong to the arrow keys (left goes right and right goes left)

sspanak:
I am glad you resolved most of your problems. It looks like we are almost good to go.

As for the ABC string - it can anything you want. It is there for user convenience, so that people know which mode is on. Usually, the first three letters of the alphabet make it clear enough. 

Now, when it comes to computers, they don't care about left-to-right or right-to-left, they just see strings. This is why you had problems with dictionary validation and this is why the ABC string appears backwards. You always have to type from left to right, so - in the dictionary, put the word first, then the frequency, and the ABC - spell it backwards. There is need to change any code in `ModeABC`.

One more thing that comes to my mind is the virtual keyboard - the keys will also be labeled backwards. `isArabic()` could be used there too.

Now let's talk about the numbers. This is something I completely forgot. Replacing the characters shouldn't be a big effort. In each mode class, in the `onNumber()` function, there is something like: `suggestions.add(String.valueOf(number));`. Instead of just using the number, there could be a `Language.getKeyNumber()`, just like there is: `language.getKeyCharacters(number)`.

However, the question is where to do it. I suppose it makes sense only in `ModePredictive` and `ModeABC`. `Mode123` is meant for typing in numeric, date, money, phone number and similar fields in apps and websites. I am certain many websites will reject Arabic numbers as invalid, so in my opinion it is best to leave Western numbers there. However, I am not familiar with the Arabic part of the Internet. If you can confirm Arabic numbers work, I am OK with leaving them in `Mode123`, too. One would have to use another language to type Western numbers.

languages/definitions/Arabic.yml
```diff
@@ -0,0 +1,16 @@
+---
+locale: ar-JO
+dictionaryFile: ar-utf8.csv
+abcString: أﺏﺕ
+hasUpperCase: no
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION_AR] # 1
+  - [ب,ت,ة,ث] # 2
```
sspanak, line 9: The letters probably need to be listed backwards, so that they would appear correctly on the screen. Must double check this before merging.

ghost: i have done some testing with this exact file in the phone and it types properly with the letters listed like this, keep in mind that if you do not have arabic locale/keyboard installed on your computer, depending on your text editor it could show in the reverse order, but the way that it is arranged now is working and good.

Arabi12321:
> I am glad you resolved most of your problems. It looks like we are almost good to go.
> 
> As for the ABC string - it can anything you want. It is there for user convenience, so that people know which mode is on. Usually, the first three letters of the alphabet make it clear enough.
> 
> Now, when it comes to computers, they don't care about left-to-right or right-to-left, they just see strings. This is why you had problems with dictionary validation and this is why the ABC string appears backwards. You always have to type from left to right, so - in the dictionary, put the word first, then the frequency, and the ABC - spell it backwards. There is need to change any code in `Language.getABCString()`.
> 
> One more thing that comes to my mind is the virtual keyboard - the keys will also be labeled backwards. `isArabic()` could be used there too.
> 
> Now let's talk about the numbers. This is something I completely forgot. Replacing the characters shouldn't be a big effort. In each mode class, in the `onNumber()` function, there is something like: `suggestions.add(String.valueOf(number));`. Instead of just using the number, there could be a `Language.getKeyNumber()`, just like there is: `language.getKeyCharacters(number)`.
> 
> However, the question is where to do it. I suppose it makes sense only in `ModePredictive` and `ModeABC`. `Mode123` is meant for typing in numeric, date, money, phone number and similar fields in apps and websites. I am certain many websites will reject Arabic numbers as invalid, so in my opinion it is best to leave Western numbers there. However, I am not familiar with the Arabic part of the Internet. If you can confirm Arabic numbers work, I am OK with leaving them in `Mode123`, too. One would have to use another language to type Western numbers.

yea thats what i thought too, i have the abcString set in the yml file just like how it is done in the hebrew and yiddish files but every time i compiled and pushed to my phone, it would just default to using the first three letters on the 2 key. This is the only reason that i made the change to the Language.getABCString() function, just for testing purposes to see whether or not the abcString would change. As it stands right now, the abcString that i have in the yml file keeps getting seen as null by the language file for some reason so the only way for the abcString to show up is for the getABCString() changes. I like your idea of using isArabic() in the virtual keyboard to display the keys from right to left but to be honest its not that important of a feature, mostly just cosmetic polish. About the numbers, I agree with you that the arabic numbers should only show up in modePredictive and ModeABC and leave Mode123 with western numbers. Most well programmed websites/apps will accept arabic numbers just the same as western numbers, even most dialers wont complain, but for the rest it is a 50/50 whether or not arabic numbers will cause something to break, even most smartphone arabic ime's include both arabic and western numbers for this reason.
I still havent been able to figure out what is causing the arabic yml abcString to be seen as null but once i figure that out i will revert the changes in Language.getABCString() back to original. Finally, my semester has started and I will probably have a little bit less time to work on this but please feel free to suggest or let me know of any changes that need to be made or that will make it better.

languages/definitions/Arabic.yml
ghost: image
image

Even the online file editor in github is inconsistent with showing the arabic from right to left or left to right (the second image with the file viewer in edit mode is correct) but this is the format that works correctly to have accurate input on the phone.

Arabi12321:
Thank you for telling me where to add support for the arabic numbers, full arabic number support has now been implemented in ModeABC and ModePredictive, tested and working. Mode123 i left untouched like you said for better support with allowing western number input into various applications/websites so as not to cause issues.

I still can't figure out what is causing the abcString from the arabic.yml to be read by Language.java as null, this is now the only outstanding issue to my knowledge.

After some testing on my phone, the ui will change the letter suggestions from left to right or right to left depending on the system locale, this is fine but it means that if we wanted to change how the suggestions and maybe even the on screen keyboard show, we might have to detect what the system locale is, or force it to be one way no matter what the android locale is set to. But to me this does not bother me as it is mostly cosmetic and doesnt affect function (i havent done any testing on the on screen keyboard however).

Arabi12321:
Ok i just took a look at the virtual keyboard (see attached photos) and i am having four issues, first the question mark is not the arabic question mark but this should be a very easy fix, Second is that all the letters displayed below the numbers are connecting as if they were trying to form a word, maybe theres a way to slightly space them out so that they show as individual letters and not connecting? third is that even in modePredictive and ModeABC, the numbers are not displaying as arabic numbers, although they type as arabic numbers (this should be the easiest to fix) and lastly in arabic locale systemwide, the ui flips itself entirely, showing 321 instead of 123 and so on, not sure how to fix this one yet.
image
image

sspanak:
OK, seeing your screenshots, I agree some more small adjustments are necessary. I believe "JO" shouldn't be displayed like this in ABC mode. The string should be the equivalent of "AR / alif-ba-ta", where "AR" is the abbreviated name of the language, not the name of the country (I suppose "JO" is for "Jordan"?). If you could provide the string I'll fix it. I can also fix the other small things.

Having said all this, I am currently busy optimizing the database, and I will have to take of several other languages, so it will probably take at least a month, until I am to finalize and release this PR. Just so you know.

Arabi12321:
yea of course, idk why its taking the second half of the locale ar-JO (yes, JO is Jordan) to display the language, but looking at other locales most are just one segment (eg "en") or both segments are the same. ive been using "أﺏﺕ" as the alif ba ta string so it should be something like "AR / أﺏﺕ".

No worries about the time frame, like I said earlier im also having more stuff to do this time as well. As long as incrementally things are getting fixed, it doesnt really matter as much how long it takes.

sspanak:
> yea of course, idk why its taking the second half of the locale ar-JO (yes, JO is Jordan) to display the language, but looking at other locales most are just one segment (eg "en") or both segments are the same.

Apologies, I checked and it seems that Java does not support neutral Arabic locale. Let's keep "ar-JO".

> i pulled out the old samsung again to reference a few things and i noticed that the equivalent of the abcString used there is just alif ba "اب" to prevent letters from connecting (ابت doesnt look as good as أﺏﺕ, but the latter uses a unicode non connecting standalone ba letter not found on the keyboard so if the أﺏﺕ doesnt work we can just use اب)

It's OK to use two letters only, as long as it is clear what it means. And I believe it is. As I said it is just a label for user convenience. The application itself doesn't care about it.

> Also is there a way to replace text with a predefined word once the correct letters have been typed, specifically replacing it with the correct harakat for certain words?

Probably, you could use the "add word" function, but I am not sure what exactly do you mean. Could you please provide an example?

sspanak:
> Nearly all arabic input methods automatically place the correct accents on الله once all the letters are typed out, so i was wondering if there is a way to automatically switch the word with this version either once the ه is typed or once the user presses space both in ModeABC and ModePredictive.

I had educate myself a bit, to understand what you mean.

Answering your question - no, unfortunately, TT9 is not able to do such complex conversions. But it isn't really necessary. It is only a matter of having the correct words in the dictionary.

For example, in French, "élève" could be written without diactritics as "eleve", but this is actually a different word. TT9 works this out by having the correct form of word in the dictionary, "élève"; and all the possible mappings "é", "è" and "e" on 3-key in the French.yml file. This way it knows that typing "35383" must yield "élève".

And as I understand, the currently included word list contains no diacritics, so you are seeing ambiguous suggestions. Can they be added somehow? Or can you find a word list the already contains them? 

Adding diacritics on-the-fly is beyond my knowledge and amount of free time, so I can not offer this solution.

sspanak:
> For this word the wordlist does include its accents on it, i was wondering if there was a way to do it in ModeABC just for this one word however, not necessarily on the fly diacritic adding for all letter, that would be way out of reach.

I see. OK, then please add the accented letters next to the regular ones in the `.yml` file. Also, if it is just one word, you could replace it with the accented variant in the dictionary file. This way it will appear correctly in Predictive mode, too.

sspanak:
> Is there anything else we need to do before merging currently?

I don't think I need anything else from you. 

As for me, after I release v27.0, I will review the code again and double check the ABC string, the text justification and if Backspace looks and works correctly (and anything else that we've discussed before); after that I'll fix any outstanding issues, then merge it.

sspanak:
@Arabi12321, I am closing this and opening a new PR from a new branch, because I can't edit your `master`.

Let's continue in #431.

PR #425, opened 2024-02-01
Backspace Handling

alexknop:
Problem:

On my QIN F21 Pro, which has a very fast processor for this application, I was noticing that at times when I press backspace/back (one button with both functions), the execution of a backspace was sometimes skipped and instead was being processed as back depending on how fast I was typing.
For instance, if I type out a sequence to start building out a word and then rapidly alternate presses between another number and backspace, it would immediately register "back" and I would get a popup on my screen (are you sure you'd like to discard this message?).

I did some debugging and found that in some instances, _isBackspaceHandled_ was false during the onKeyUp event of pressing backspace. Chances are it was failing some checks during onBackspace() because it always had the same keyCode which indicates a backspace.

Solution:

Remove the _isBackspaceHandled_ variable, instead check whether the key is backspace and if so, continue to run onBackspace in the onKeyDown function, but when it comes to onKeyUp, just check whether it is backspace and if so return true.

I have tested on my F21 Pro and it has solved this issue. I also tested with a slower Android 8.1 device and found no issue with this change. Holding down backspace to clear multiple letters works still.

### DIFF
--- a/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
+++ b/src/io/github/sspanak/tt9/ime/KeyPadHandler.java
@@ -124,7 +124,6 @@ abstract class KeyPadHandler extends InputMethodService {
 			return true;
 		}
 
-
 		// start tracking key hold
 		if (Key.isNumber(keyCode)) {
 			event.startTracking();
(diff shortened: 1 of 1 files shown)


alexknop:
Closing for now as this isn't a foolproof solution.

Issue #137, opened 2022-12-23
arabic keyboard language


islam2hamy:
Hi ,

Can you please add arabic keyboard language

Thank you

sspanak:
No, sorry, it is beyond my capabilities. I have no idea how T9 is supposed to work in Arabic (which key should correspond to which letter and how to deal with the different letter forms); nor can I read the language to verify it actually works.

Anyway, if you speak the language, you are welcome to follow the [adding a new language guide](https://github.com/sspanak/tt9/blob/master/CONTRIBUTING.md#adding-a-new-language). If you have at least basic programming skills, it should be fairly easy to add it or at least determine whether the typing engine can support Arabic (I am not even sure about that).

sspanak:
I am unsure why this is closed, but Arabic is certainly not included, so I am reopening it.

Having gained some experience with Hebrew, I think adding Arabic will not be so difficult. At this point I'll need only the following:

1. An example layout (what letter goes on what key), including any special punctuation marks that need to be included or removed from 1-key and 0-key.
2. A good word list containing Arabic words, personal names, names of places and whatnot. Anything that one would type daily.

islam2hamy:
Hello @sspanak ,

please take a look

1. 

2. https://onattech.github.io/Arabic-Keyboard/

sspanak:
Well... I meant a 12-key layout, not a QWERTY layout. You know, which letters are supposed to appear when you press "2", "3", "4", ...; and if there is need to add extra punctuation characters (non-standard to Latin/Cyrillic alphabets).

I really don't know what to do with a computer keyboard layout. :smile:

islam2hamy:
@sspanak , sorry for being late

sspanak:
Sorry, that photo didn't load the last time I opened Github. I see it now. 

However, I was hoping that you (or someone else for that matter) could actually write the appropriate letters/characters for each key. See, I can't read Arabic and it is very difficult for me to compare the letters on your photo with the Wikipedia article about Arabic and figure out the correct ones for 2-key, 3-key and so on, especially given the fact that the font used on your photo is kind of "square", which makes the visual comparison even harder.

islam2hamy:
> Sorry, that photo didn't load the last time I opened Github. I see it now.
> 

No , it is my fault , uploading the photo took time from me

[...]

PR #431, opened 2024-02-10
Arabic

sspanak:
Moved the code from #412 to a new branch, because I can't edit @Arabi12321 `master` branch.


sspanak:
> I have just finished testing your apk and everything is working exactly as it should be, all the issues that we mentioned earlier are now solved. The on screen keyboard is accurate now, abcString is working properly, letters are mapped correctly and the arabic numbers are showing up correctly as well. I see the new frequency list in the edits and now the dictionary typing has improved its predictive accuracy significantly than before. So now everything is well with this apk.

See my answer in #137

PR #436, opened 2024-02-17
Adding emoji

### LINKED ISSUE #363: Custom emojis without requiring to custom compile
Hi !
Loving the work you've done here and would like to start by thanking you. Truly a life savior !

I'm in the process of switching from my smartphone to a Qin F22 pro and your keyboard really shines in those dark times. The only """"issue"""" I encountered so far, is that the emoji list doesn't suite my taste nor my preferences.
I'm a heavy user of 🙇, 😊 and other non-present emojis. 

Can you think of an idea on how to make this custom ? 
I was thinking of a setting where one could set its custom list of emojis, with just a few spots, like 5-10 emojis top, from a list of available emojis. The custom list would come first when choosing emojis, then positive, negative, etc...

I did not, yet, found time to look at the code base, but would love to contribute with your guidance.

sspanak:
Resolves #363

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/db/DictionaryLoader.java
+++ b/app/src/main/java/io/github/sspanak/tt9/db/DictionaryLoader.java
@@ -21,6 +21,7 @@ import io.github.sspanak.tt9.db.sqlite.InsertOps;
 import io.github.sspanak.tt9.db.sqlite.SQLiteOpener;
 import io.github.sspanak.tt9.db.sqlite.Tables;
 import io.github.sspanak.tt9.ime.TraditionalT9;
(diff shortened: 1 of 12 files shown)


sspanak:
bug 1:
1. Select 123
2. Type a currency
3. Switch to ABC
4. Type "space"

result: Pressing "0" has no effect

bug 2:
1. Select ABC
2. Type a currency
3. Try to switch to the next mode

result: switching does not work until one of 1-9 is pressed

sspanak:
[emoji can now be added as custom words](https://github.com/sspanak/tt9/pull/436/commits/a0cc47145547d2c9a904ce1d3d876853e0bfe741) is the first commit here.

PR #437, opened 2024-02-19
Database migrations

### LINKED ISSUE #381: Automatically load dictionary on initial use
Hi, thanks for this amazing project.

I think it would be very useful if we can get the Dictionary automatically loaded on initial use.

While I am not too good in Java development, I would think the easiest unprofessional way would be as follows:

1. When IME detects there are no words in the database (which currently triggers a toast notification), it should temporarily set a boolean noDict=true
2. Thereafter, it should open the Settings page
3. On starting the Settings page, it checks noDict's value, if true, then [open the Languages section, and] set noDict=false & trigger Load dictionary.

@sspanak Please let me know your thoughts, and once again, thanks very much for all your work!

sspanak:
Resolves #381


sspanak:
Must update the translations and then it should be good to go.

PR #438, opened 2024-02-22
Portuguese

sspanak:


docs/dictionaries/ptWordListReadme.txt
```diff
@@ -0,0 +1,10 @@
+Wordlist:
```
sspanak: The information here is incomplete

app/languages/definitions/Portuguese.yml
```diff
@@ -0,0 +1,14 @@
+locale: pt-PT
+dictionaryFile: pt-PT-utf8.csv
```
sspanak, line 2: The file contains invalid words.

PR #440, opened 2024-02-23
Updated Ukrainian translation

WallK:
Also hanged locale name to be "ua"


WallK:
Previous translation had a lot of errors and looked like a translation from Russian

WallK:
Also, why are some things non-translatable?
For example "Cannot add a word when no language is selected." is marked as non-translatable

WallK:
Sure, will do

res/values-ua/strings.xml
```diff
@@ -0,0 +1,114 @@
...
+	<string name="add_word_invalid_language" translatable="false">Cannot add a word when no language is selected.</string>
+	<string name="add_word_success">\"%1$s\" додано.</string>
+	<string name="add_word_title">Додати слово</string>
+
+	<string name="pref_category_about">Про програму</string>
+	<string name="pref_category_abc_mode">Режим АБВ</string>
+	<string name="pref_category_hacks">Сумістність</string>
+	<string name="pref_category_appearance">Зовнішній вигляд</string>
+	<string name="pref_category_debug_options" translatable="false">Debug Options</string>
+	<string name="pref_category_predictive_mode">Предективний режим</string>
```
sspanak:
```suggestion
	<string name="pref_category_predictive_mode">Предиктивний режим</string>
```

Isn't this a spelling mistake?

res/values-ua/strings.xml
```diff
@@ -0,0 +1,114 @@
...
+	<string name="setup_default_keyboard">Вибрати стандартну клавіатуру</string>
+	<string name="setup_tt9_on">Traditional T9 увімкнено</string>
+	<string name="setup_tt9_off">Traditional T9 вимкнуто</string>
+	<string name="setup_click_here_to_enable">Нажміть тут, щоб увімкнути TT9 в налаштуваннях Android.</string>
+
+	<string name="key_hold_key">(тримати)</string>
+	<string name="key_none" translatable="false">--</string>
+	<string name="key_back">Назад</string>
+	<string name="key_call">Подзвонити</string>
+	<string name="key_dpad_up" translatable="false">▲</string>
```
sspanak: Please, remove the non-translatable strings. TT9 will read them from the base English strings.xml. The special characters are the same in all languages, so it makes no sense to duplicate them.

res/values-ua/strings.xml
```diff
@@ -0,0 +1,114 @@
...
+	<string name="pref_category_about">Про програму</string>
+	<string name="pref_category_abc_mode">Режим АБВ</string>
+	<string name="pref_category_hacks">Сумістність</string>
+	<string name="pref_category_appearance">Зовнішній вигляд</string>
+	<string name="pref_category_debug_options" translatable="false">Debug Options</string>
+	<string name="pref_category_predictive_mode">Предективний режим</string>
+	<string name="pref_category_function_keys">Гарячі клавіші</string>
+	<string name="pref_category_keypad">Клавіатура</string>
+	<string name="pref_category_setup">Початкові налаштування</string>
+	<string name="pref_category_usage_stats" translatable="false">Usage Stats</string>
```
sspanak: This is is used only in development mode, there is no need to translate it. Under normal circumstances the regular users shouldn't care about it. This applies to all sentences marked as `translatable="false"`

res/values-ua/strings.xml
```diff
@@ -0,0 +1,114 @@
...
+	<string name="pref_hack_fb_messenger_summary">Дозволяє відправляти повідомлення по натиску OK в Facebook Messenger.</string>
+	<string name="pref_show_soft_function_keys">Показувати кнопки на екрані</string>
+	<string name="pref_show_soft_numpad">Показувати цифрову клавіатуру на екрані</string>
+	<string name="pref_show_soft_numpad_summary" translatable="false">(BETA)</string>
+	<string name="pref_help">Допомога</string>
+	<string name="pref_upside_down_keys">Зворотній порядок клавіш</string>
+	<string name="pref_upside_down_keys_summary">Використовуйте це налаштування, якщо у вас в першому ряді 7–8–9 замість 1–2–3.</string>
+
+	<string name="dictionary_cancel_load">Скасувати завантаження</string>
+	<string name="dictionary_load_bad_char">Завантаження не вдалося. Невірне слово \"%1$s\" у рядку %2$d мови \"%3$s\".</string>
```
sspanak:
```suggestion
	<string name="dictionary_load_bad_char">Завантаження не вдалося. Невірне слово «%1$s» у рядку %2$d мови «%3$s».</string>
```

Isn't it nicer to use the Ukrainian/French quotes instead of English ones? As I understand this is the standard.

sspanak (changes requested):

res/values-ua/strings.xml
WallK: Yeah, it is, thanks

res/values-ua/strings.xml
WallK: I just took English xml and translated in-place
Will clean up untranslatable stuff and there's less of it now (:

res/values-ua/strings.xml
WallK: I don't agree at this point
Usually, double quotes are nicer in the interfaces and have much more font support

But it is technically true that we use those Christmas tree quotes, so the decision is up to you

sspanak: It's all the same to me. I just thought this is the Ukrainian way and people would like it better when they see properly formatted text. But if you guys don't care either, let's leave the English quotes.

WallK:
Do you want me rebasing to version-28 and putting new translation in the \app\src\main\res\values-ua\ ?

sspanak:
> Do you want me rebasing to version-28 and putting new translation in the \app\src\main\res\values-ua\ ?

Yes, this is the correct path now.

sspanak:
@WallK, I am not sure if you want to continue working in this PR or open a new one, but whatever you choose, could you also change the "" key to "Key Repeat Protection" or "Accidental Key Repeat Protection", whatever sounds better in Ukrainian? I will change it in all other languages. 

Sorry for the last minute change, but I think this is a bit more clear than "Multi-press protection".

WallK:
Rebase closed the PR automatically (???)
I already translated something along the lines of "Accidental Key Repeat Protection" because "Multi-press protection" didn't make a nice sentence in Ukrainian, glad we thought alike

app/src/main/res/values-ua/strings.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_settings">Налаштування TT9 Settings</string>
```
sspanak:
```suggestion
    <string name="app_settings">Налаштування TT9</string>
```

sspanak (approved):

WallK:
Ok, I was to hasty and the real language code is "uk", sorry
I guess you should rename the folder back...

PR #441, opened 2024-02-23
Docs 28

sspanak:


WallK:
About "Download Dictionary" to "Завантажити Словник / словник"
I'm not sure about capitalization here, it felt more of THE dictionary here, that's why I capitalized it specifically
But it is inconsistent with other strings, thank you

sspanak:
> About "Download Dictionary" to "Завантажити Словник / словник" I'm not sure about capitalization here, it felt more of THE dictionary here, that's why I capitalized it specifically But it is inconsistent with other strings, thank you

No big deal. It means "load what you have just selected". There is no accent on "dictionary". The English version is capitalized, because the grammar rules in English require all words in a title to be capitalized.

Issue #393, opened 2023-11-14
Center OK Button Issues


theancientways:
Hey, I'm encountering an issue on the Sonim XP3plus where the center ok button doesn't work sometimes in the contacts app. The only time it consistently doesn't work is when I try to use the "add a contact" function from the text message app. It takes me to a window with a search field above a list of contacts, the top option of which being "create a contact". This is the window where the center button consistently doesn't work. It does work if I switch to the phone's native text input method. I know you didn't test this on every phone so there's bound to be bugs here an there on different devices but still thought I'd report it.

theancientways:
Some more info... Oddly, after toggling off the "send messages with OK in messenger", the center button seems to function in the contacts app only once, after that it doesn't work. Weird.

sspanak:
I don't quite get what's wrong, because you just say "it doesn't work", but I sounds really similar to [this bug](https://github.com/sspanak/tt9/pull/364). If it is the same, it will be fixed in the next version. Or is it different?

theancientways:
Hey, I don't know if I have the know-how to be able to explain it beyond how I did in my first post, basically in a nutshell the center OK button on my flip phone doesn't work when a new number texts me and I try to add them as a new contact. From the text message thread I press a soft key and select "add new contact" from the popup list. That takes me to the contacts app with a button that says "add new contact", trying to press this with the OK button does nothing. Sorry I can't explain better but I could get screenshots if that might help. It's not the worst of inconveniences but it's just odd that it doesn't work right.

If I download your new release, will it carry over my user dictionary or will I have to start it from scratch?

Thanks as always.

theancientways:
hey just a bump for my question at the end of my above post, is there a way to save your user dictionary when you upgrade or do you have to start from scratch?

sspanak:
This is quite problematic at the moment. The upcoming version 27.0 will introduce major database changes, which will make it possible to protect the added words. It would even be possible to delete all languages, then reload them, but the words will remain safe. In the future, it would be possible to export them, as well.

However, the above comes at the cost of deleting absolutely all words when upgrading to v27.0.

But still, you may actually win more than you lose after the upgrade, depending on the language(s) you are using. Bulgarian, English, Romanian, Russian and Ukrainian will receive a major update including thousands of new words.

I realize many other languages lack rich dictionaries, but hopefully, with community contributions, it will get better over time.

theancientways:
[...]

PR #475, opened 2024-03-22
Kyocera Default 'Message' App Support

owentroe:
This MR fixes behavior on the Kyocera DuraXA flip phone where sending a message with `DPAD_CENTER` results in a double send attempt, resulting in an error screen after each message is sent. 

A new app hack has been added to only send the `KEYCODE_DPAD_CENTER` event, and not attempt the `KEYCODE_ENTER` code that is present in `onEnterDefault()`. A toggle switch inside of the 'Initial Setup' menu has been added. 

I did not add menu strings for other (non-english) locales.

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/ime/helpers/AppHacks.java
+++ b/app/src/main/java/io/github/sspanak/tt9/ime/helpers/AppHacks.java
@@ -58,6 +58,18 @@ public class AppHacks {
 		);
 	}
 
+	/**
+	 * isAndroidMMS
+	 * Android Messages on Kyocera devices will register the OK key as both ENTER and DPAD_CENTER when sending a message using TT9. This results
+	 * in an attempt to double send causing an 'Empty Message' error page to pop up. 
+	 */
+	private boolean isAndroidMMS() {
+		return isAppField(
+			"com.android.mms",
+			EditorInfo.TYPE_CLASS_TEXT | EditorInfo.TYPE_TEXT_FLAG_MULTI_LINE | EditorInfo.TYPE_TEXT_FLAG_CAP_SENTENCES
+		);
+	}
+
 
 	private boolean isGoogleChat() {
 		return isAppField(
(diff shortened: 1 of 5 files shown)


sspanak:
Thanks for your contribution! Anyway, I think it may not be needed. I believe the problem is related to #393. I would like to fix that first and possibly reduce the number of hacks, including the double keycode sending, which causes problems on your phone.

Would it be OK for you to test the fix for #393 once it is out, and only if you are still experiencing problems, then I'll consider merging your PR?

owentroe:
> Would it be OK for you to test the fix for #393 once it is out, and only if you are still experiencing problems, then I'll consider merging your PR?

Yeah, feel free to notify via this PR when ready. I didn't see an associated PR for issue #393, otherwise I could build and test sooner.

sspanak:
> > Would it be OK for you to test the fix for #393 once it is out, and only if you are still experiencing problems, then I'll consider merging your PR?
> 
> Yeah, feel free to notify via this PR when ready. I didn't see an associated PR for issue #393, otherwise I could build and test sooner.

There isn't one, but I am planing to fix that in the next release.

sspanak:
@owentroe, my attempt to fix the bug is ready to test. Could you please try [v29.15](https://github.com/sspanak/tt9/releases/download/v1.0-inital/tt9-v29.15.apk). Alternatively, if you prefer building the APK on your own, just use the `redo-ok` branch.

Any feedback will be much appreciated!

sspanak:
@owentroe, I have done multiple improvements for the OK key, which I believe, have fixed the problem. The new [v30.0](https://github.com/sspanak/tt9/releases/tag/v30.0) is officially out. Could you please try it out and report back if you are still having any issues?

owentroe:
@sspanak I just tested this (or at least the latest release in F droid), and the issue in the Kyocera messaging app still persists.

owentroe:
I managed to load v30, looks like it's working now (no double send bug). I will close this PR.

PR #484, opened 2024-04-04
Add Croatian language and dictionary

danijelt:
Dictionary taken from Openboard and cleaned up from non-Croatian words.

### DIFF
--- a/app/languages/definitions/Croatian.yml
+++ b/app/languages/definitions/Croatian.yml
@@ -1,0 +1,13 @@
+locale: hr
+dictionaryFile: hr-utf8.csv
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION] # 1
+  - [a, b, c, č, ć] # 2
+  - [d, e, f, đ] # 3
+  - [g, h, i] # 4
+  - [j, k, l] # 5
+  - [m, n, o] # 6
+  - [p, q, r, s, š] # 7
+  - [t, u, v] # 8
+  - [w, x, y, z, ž] # 9
--- a/docs/dictionaries/hrWordListReadme.txt
+++ b/docs/dictionaries/hrWordListReadme.txt
@@ -1,0 +1,9 @@
+Croatian wordlist 1 by: Openboard
+Version: 7bf52c1 (2020-01-15)
+Source: https://github.com/openboard-team/openboard
+License: https://github.com/openboard-team/openboard/blob/master/LICENSE
+
+Croatian wordlist 2 by: Goran Igaly
+Version: 94b795d (2022-02-05)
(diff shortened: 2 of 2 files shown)


sspanak:
Great! Thanks for adding one more language!

I can see the dictionary consists of 209k words. Based on my experience with other Slavic languages, predictions will be ok-ish, most of the time, but every day, you will find a word or two are missing. I've used Bulgarian with 256k words for some time, so I am really speaking from personal experience.

Do you have some institute that regulates the language, or some respected university that has published something like "the big dictionary of Croatian"? Almost any country has one of these, and usually, they are much better compared to the Google dictionary, generated from subtitles, that you have added. They are spell-checked and contain huge amount of words, sometimes, over 1 million. If you know of such dictionary, I would be happy to include it. It will improve the typing experience immensely. 

Otherwise, we can go with this one. I can add some more words from other sources that I know of, but as I said, it won't be optimal.

danijelt:
This wordlist is poor, I know, but it's the best one I could find that includes the word frequency.

There's one big list of 624k words that includes inflections, but it doesn't list the frequency: https://github.com/gigaly/rjecnik-hrvatskih-jezika

If you think such a merged list with and without frequency would be fine, I can probably find some time to do it in the next few days.

sspanak:
Perfect! This is what I am talking about. I'll can add the frequencies to the larger one, I have a handy script for that, so don't worry about it. I'll have to do some extra processing for faster performance anyway.

Just fyi, the frequencies are not that important. Even if only 20% of the words have frequencies, the predictions will still be very good.

I have some more work on the next release, but after that, I believe, I can include Croatian too. Meanwhile, if you find even a bigger word list, you are welcome to share it.

danijelt:
Updated with merged dictionaries, with a total of 651915 words.

sspanak:
Great, thank you! When I get to this, I'll do some more processing and will most probably include it in the next release.

sspanak:
@danijelt, I have reviewed the words. It seems that there are a lot of chemistry and math terms. Probably, they were collected from science articles. This is nice, but I hope there are enough common everyday words for nicer experience.

The second problem I noticed is, some of the words are weirdly spelled, for example: MOLDOVE, DVADESETŠESTA, FOTOKEMIJSKO, KUFER, TABLO. Some others were in lowercase, but had 1-2 uppercase letters at the beginning. I didn't write down any examples, though. Anyway, I tried to fix as many as possible, but I have probably missed some.

In both cases, if you notice any incorrect spellings or nonsense, you are welcome to open a new issue or even a PR.

danijelt:
The author of the dictionary works at the Faculty of Science, so that's probably it. I'll use it as-is now and I might try sourcing new words elsewhere if this list won't be enough.

PR #509, opened 2024-05-05
Add Hungarian dictionary

gyadam:
Not sure if this is the best source for words, but I needed something to be able to text faster on my dumb phone. I've been using it for the past week and it works great for me (I did have to clean up the word list a little bit to build an APK, so it's not the exact same one as the source).

Note that I've added letters with special characters in between the normal letters, e.g. `aábc`. This is intentional as it's how I remember old dumb phones working with the Hungarian alphabet and it's logical since we use a lot of these letters.

### DIFF
--- a/app/languages/definitions/Hungarian.yml
+++ b/app/languages/definitions/Hungarian.yml
@@ -1,0 +1,13 @@
+locale: hu-HU
+dictionaryFile: hu-utf8.csv
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION] # 1
+  - [a, á, b, c] # 2
+  - [d, e, é, f] # 3
+  - [g, h, i, í] # 4
+  - [j, k, l] # 5
+  - [m, n, o, ó, ö, ő] # 6
+  - [p, q, r, s] # 7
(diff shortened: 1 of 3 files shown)


sspanak:
Thanks for contributing to the project! 

The word list looks mostly OK. There some weird word such as: "fellépésfellépésfellépésfellépés", "uunndddooorrríííííttttttttttttóóóóóó" and "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa". I'll try to clean these before merging, because they don't make much sense.

> Note that I've added letters with special characters in between the normal letters, e.g. aábc. This is intentional as it's how I remember old dumb phones working with the Hungarian alphabet and it's logical since we use a lot of these letters.

That's OK. Just have in mind the order will be preserved only in ABC mode. In Predictive mode, the letters that are used the most will be shifted to the beginning of the list.

sspanak:
@gyadam, I have cleaned up the word list. Unfortunately, I had to throw away about 30% of the words, but still it works quite well. I'll include Hungarian in the next release.

gyadam:
Awesome, thanks for cleaning it up and merging @sspanak!

PR #551, opened 2024-06-25
UI resize

sspanak:


sspanak:
The key height steps are too large. Adjust them to sane values.

sspanak:
Lock the key size setting, when the Layout != virtual keypad

PR #559, opened 2024-07-02
disabled the 'send with OK in Messenger'

sspanak:


app/src/main/res/xml/prefs_screen_setup.xml
```diff
@@ -19,6 +19,7 @@
 
 		<SwitchPreferenceCompat
 			app:defaultValue="false"
+			app:enabled="false"
```
sspanak, line 22: Update the documentation too.

PR #566, opened 2024-07-04
Copy pasta

sspanak:
- ~~In Numpad mode, disable the function keys on both sides.~~
- ~~In Numpad mode, minimizing then maximizing while the text editing tools are shown, causes the numbers to appear too.~~
- ~~In Tray mode, the height becomes incorrect after closing the palette~~
- ~~"Back" must close the text palette, but it hides the entire UI instead~~
- Make the commands work
- ~~Test both layouts in RTL mode~~
- ~~Make sure the word skipping keys work properly in RTL mode~~
- ~~Translations~~
- ~~Sort out command key 3 and command key 5 visibility when the respective command is not supported~~
- ~~Using simulated DEL key to clear the text on paste causes a race condition, which causes the last character of the pasted text to be erased (sometimes). Probably, it is a bad idea to simulate the key press on cut too.~~
- ~~Add a "back" button for the virtual keyboard. Firefox takes over the system "back", so it we can't go back to the numeric keys.~~


abea210:
Hi
I went through the threads several times regarding the copy paste,
And I didn't quite understand,
I want to know if there is already such an option or if it is still pending,
 It will help me a lot

Thank you

abea210:
> > Hi I went through the threads several times regarding the copy paste, And I didn't quite understand, I want to know if there is already such an option or if it is still pending, It will help me a lot
> > Thank you
> 
> Development is still in progress.

Thank you
Is it something that will probably take a few days/weeks, or something that will take a few months?

sspanak:
> > > Hi I went through the threads several times regarding the copy paste, And I didn't quite understand, I want to know if there is already such an option or if it is still pending, It will help me a lot
> > > Thank you
> > 
> > 
> > Development is still in progress.
> 
> Thank you Is it something that will probably take a few days/weeks, or something that will take a few months?

It is done when it is done. I can not guarantee about deadlines.

PR #578, opened 2024-07-18
Improvements 36

sspanak:


app/src/main/java/io/github/sspanak/tt9/ui/main/BaseMainLayout.java
```diff
@@ -47,7 +49,11 @@ void setDarkTheme(boolean dark) {}
 
 	protected View getView() {
 		if (view == null) {
-			view = View.inflate(tt9.getApplicationContext(), xml, null);
+			// Adding the ContextThemeWrapper fixes this error log:
+			// "View class SoftKeyXXX is an AppCompat widget that can only be used with a
+			// Theme.AppCompat theme (or descendant)."
+			ContextThemeWrapper themedCtx = new ContextThemeWrapper(tt9.getApplicationContext(), R.style.TT9Theme);
```
sspanak: This causes the key ripple effect to use the system color, instead of the TT9 color. Must fix it.

sspanak: Actually, every phone works a bit differently. It is fine.

PR #591, opened 2024-08-02
Small fixes 37

sspanak:


sspanak:
Must add a "large" layout for the InputText and NoResults items.

PR #594, opened 2024-08-04
Add support for Catalan.

### LINKED ISSUE #592: Support for Catalan Language
I was wondering if you could add support for the Catalan language. I have been using this app for some days now an if has been flawless for the English and Spanish! The the only thing I'm missing is to be able to type in native language (Catalan).

It's pretty similar to Spanish but it has apostrophes both at the start of words (l'aigua) and at the end (Dona-t’ho) as well as - in the middle (anti-inflamatori).

Possible Keys (adapted from Portuguese):
```locale: ca-ES
dictionaryFile: ca-ES-utf8.csv
name: Català / CA
layout:
  - [SPECIAL] # 0
  - [PUNCTUATION] # 1
  - [a, b, c, ç, à] # 2
  - [d, e, f, è, é] # 3
  - [g, h, i, í, ï] # 4
  - [j, k, l] # 5
  - [m, n, o, ò, ó] # 6
  - [p, q, r, s] # 7
  - [t, u, v, ú, ü] # 8
  - [w, x, y, z] # 9
```
We also have l·l (Col·lecció), I don't know if that should be in the same key as l (which would be the best option in my opinion, even if theoretically it's 3 keys long, as it is pronounced pretty much the same as a single l).

As for punctuation, it's the same as Spanish but without ¡ and ¿, so I guess PUNCTUATION_FR?

I have been looking for good dictionaries for Catalan but I have not had much success. Here are the best two I have found:
[This one is the most promising](https://codeberg.org/Helium314/aosp-dictionaries/src/branch/main/wordlists_experimental/main_ca.combined.gz). It has also listed the most common words after each word (which is interesting but I don't think it will be usefult in our case) and has words that contain - or ' listed as a whole word (I don't know if it will be useful, but I guess it is better than nothing). There are other dictionaries (both Catalan and other languages) in that repository, so take a look if this one is not useful for our use-case. 
[And this one](https://github.com/jig/diccionari/blob/master/catala.dic). Which has - and ' words listed as a whole new word.
Sadly, I have been unable to find an official dictionary.

I can also provide both menu and documentation translations for both Catalan and Spanish.

Thank you very much for providing a keyboard as good as this one free of charge! I'll make sure to make a donation when I can spare some money!

Let me know if you need any extra information!

Roconx:
Mostly fixes #592.

Missing support for l·l, but I don't think it is that important.

A way to continue recommending characters when words have - or ' would be nice, would also fix the issue with l·l.


app/languages/definitions/Catalan.yml
```diff
@@ -0,0 +1,14 @@
+locale: ca-ES
+dictionaryFile: ca-ES-utf8.csv
```
sspanak: `ca-utf8.csv` should be fine. In Portuguese, it makes sense to add the country code, because the European and the Brazilian variants are way too different. So, unless you are aiming for very specific Spanish-only Catalan, the country code can go away.

app/languages/definitions/Catalan.yml
```diff
@@ -0,0 +1,14 @@
...
+name: Català
```
sspanak: TT9 knows how to generate the name by using the locale. Specifying this property is only necessary, if it appears incorrectly in the language list. Please, try without a `name`.

app/languages/definitions/Catalan.yml
```diff
@@ -0,0 +1,14 @@
...
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION_FR, ·] # 1
```
sspanak, line 5: Let's remove the dot and add the [l-dot](https://www.compart.com/en/unicode/U+0140) character on the 5-key. As I explained in #592, combining characters are potential trouble makers.

sspanak: Scratch that. As per the following comments, the middle dot will live separately.

sspanak:
@Roconx, I have re-added the large wordlist and cleaned it up as much as possible. After that I did a quick typing test by using 6 Wikipedia articles and it was mostly fine, but I am under the impression many words are missing. This is surprising, given that the dictionary now contains more than 900k words. Or maybe I picked too specialized articles, I don't know.

Please, try the latest commit from this branch and let me know if there is anything weird. 

Also, please let me know which of the following contractions/combinations are possible in Catalan:
```
'en
'hi
'ho
'l
'ls
'm
'n
'ns
's
't
-d'
-de-
-el
-els
-em
-en
-en-
-ens
-hi
-ho
-i-
-in
-l
-la
-la-
-le-
-les
-li
-lo
-los
-m
-me
-n
-ne
-ne-
-nos
-out
-s
-se
-t
-te
-us
-vos
-vu
```

Some of them come from French words, but I wasn't sure if they are also used in Catalan, so I took a conservative approach and kept them.

If you don't have any comments, I'll just merge the branch as it is.

Roconx:
Thank you very much for doing this! I wanted to do it but have been very busy lately. I'll take a look tomorrow and I'll let you know!

Roconx:
Wow, you were pretty much spot on! I managed to find a pretty detailed source [here](https://www.auladecatala.com/ortografia-dels-pronoms-febles/) and I think that this is the final list (feel free to check in case I missed any): 
```
'hi
'ho
l'
'ls
'ls-
'm
'n
n'
'ns
'ns-
's
-t'
d'
m'
t'
s'
-el
-els
-em
-en
-ens
-hi
-ho
-l
-la
-les
-li
-li'
-li-
-lo
-los
-los-
-m
-me
me'
-me'
-me-
-nos
-nos-
-se
-se'
-t
-te
-te'
-te-
-us
-us-
-vos
-vos-
-m'
```

sspanak:
Alright, I think we sorted everything out. I am merging this PR and will include Catalan in the next version. If you notice any misspelled words, feel free to open an issue or a PR to correct them.

PR #608, opened 2024-08-26
Backspace acceleration

sspanak:


sspanak:
Messed up the history. Development continues in #612

PR #611, opened 2024-08-29
Typing Improvements 38, part 1

sspanak:
- ~~update the documentation on how to type new words~~
- ~~(bug) Pressing, then holding double function virtual keys, does not reset the repeat state. For example, filtering, then clearing the filter, then filtering again, causes double filter to be applied. Allow the secondary function to clear the "last pressed" state.~~


sspanak:
Messed up the history. Development continues in #612

PR #650, opened 2024-10-16
Small fixes 40

sspanak:


sspanak:
Regression: automatic language loading is broken.

sspanak (commented):
Regression: automatic language loading is broken.

PR #662, opened 2024-10-31
New dictionary format

sspanak:
Sonim XP3800 loading times:

| /	| English | Bulgarian | 
|---|---------|-----------|
| v40.0	| 31 s | 288 s |
| v40.5	| 27.5 s  | 206 s |

Qin F21 Pro+ loading times:

| /	| English | Bulgarian | 
|---|---------|-----------|
| v40.0	| 3 s | 24.9 s |
| v40.5	| 2.9 s  | 18.9 s |


app/build.gradle
```diff
@@ -18,34 +25,40 @@ tasks.register('validateLanguages') {
...
+tasks.register('generateDownloads') {
+	inputs.dir DICTIONARIES_INPUT_DIR
+	outputs.dir DICTIONARIES_DOWNLOAD_DIR
+
+	dependsOn buildDictionaries
+	mustRunAfter buildDictionaries
 
 	doLast {
-		convertHelpDocs(HELP_MARKDOWN_DIR, HELP_HTML_DIR)
+		createDownloads(DICTIONARIES_OUTPUT_DIR, DICTIONARIES_DOWNLOAD_DIR)
```
sspanak: - Instead of generating dictionaries assets, generate dictionary downloads
- Put somewhere the hash of the source dictionary. Only if the hash differs, then build a new dictionary.
- Always copy from download/ to full/assets/. It's a quick operation.

app/build-dictionary.gradle
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    return [errorCount, errorMsg]
+}
+
+
+//////////////////// DICTIONARY PROCESSING ////////////////////
+
+static byte[] compressDictionaryLine(String digitSequence, List<String> words) {
+    return (digitSequence + words.join()).getBytes(StandardCharsets.UTF_8)
```
sspanak: German fails because of "ß", which is typed using "77". This will be a problem in Korean too.

Solution: 
- In case of a discrepancy between the word length and the sequence length, separate the words with a space. 
- Then, when loading in `WordFile`, if there is a space, just continue the while loop that collects the word list, instead of appending the space.
- Check for `wordLength == sequenceLength` only if there was NO space.

sspanak (commented):
Verify all downloads links are valid before merging.

PR #671, opened 2024-11-22
Korean

sspanak:


app/src/main/java/io/github/sspanak/tt9/ime/CommandHandler.java
```diff
@@ -159,30 +165,37 @@ protected void nextLang() {
...
 		int previous = mEnabledLanguages.indexOf(mLanguage.getId());
 		int next = (previous + 1) % mEnabledLanguages.size();
-		mLanguage = LanguageCollection.getLanguage(getApplicationContext(), mEnabledLanguages.get(next));
+		mLanguage = LanguageCollection.getLanguage(mEnabledLanguages.get(next));
 
 		// validate and save it for the next time
 		validateLanguages();
+
+		// for languages that do not have ABC or Predictive, make sure we remain in valid state
+		mInputMode = determineInputMode();
```
sspanak:
```suggestion
```

This prevents from dynamic language switching while typing a word. Instead do this in `ModeABC.changeLanguage()`:
```java
		if (newLanguage != null && newLanguage.isSyllabary()) {
			return false;
		}
```

PR #692, opened 2024-12-26
Tamazight

sspanak:


docs/dictionaries/zghWordlistReadme.txt
```diff
@@ -0,0 +1,4 @@
+Tamazight wordlist by: Abdelhaque Id Ali
```
sspanak, line 1: Squash the last commits before merging.

app/languages/definitions/TamazightLatin.yml
```diff
@@ -0,0 +1,14 @@
+locale: zgh-DZ # zgh = Morroccan Tamazight, DZ = Algeria. We combine these just to get a unique ID.
```
sspanak, line 1: "zgh / dz" causes incorrect country code in ABC mode

sspanak: Fixed

PR #702, opened 2025-01-07
Bugs 44

### LINKED ISSUE #333: Swipe left to delete last word
On the on-screen keyboard, it would be nice if we could swipe left on the delete button to delete the last word!

Its not an essential feature, just a nice one!

### LINKED ISSUE #701: Swipe letter keys in T9 mode to make them act like ABC mode
I read in another thread that you had managed to enable swiping for all keys.

Would the following be possible?

Allow swiping letter keys in T9 mode to get individual letters. This would avoid having to switch between T9 and ABC.

For example, the 5 key, You could swipe left for 'j', swipe up for 'k' or swipe right for 'l'.

You could even arrange the key labels in a diamond shape to make this more intuitive.

Perfect for typing words that are not in the dictionary.

sspanak:
Fixes #333 
Fixes #701

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/ime/helpers/StandardInputType.java
+++ b/app/src/main/java/io/github/sspanak/tt9/ime/helpers/StandardInputType.java
@@ -112,8 +112,11 @@ abstract public class StandardInputType {
 	}
 
 
-	abstract protected boolean isDefectiveText();
+	abstract public boolean isDefectiveText();
 
+	private boolean isNoSuggestionsText() {
+		return isText() && (field.inputType & InputType.TYPE_MASK_FLAGS & InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS) == InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+	}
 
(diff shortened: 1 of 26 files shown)


sspanak:
This seems to break auto-capitalization in the Contacts on Android 14/15

sspanak:
The Contacts still work incorrectly on Android 11.

app/src/main/java/io/github/sspanak/tt9/ime/helpers/StandardInputType.java
```diff
@@ -192,14 +195,14 @@ public int determineTextCase() {
...
 
-		if (isSpecialized()) {
-			return InputMode.CASE_LOWER;
-		}
-
 		if (isPersonName()) {
 			return InputMode.CASE_CAPITALIZE;
 		}
 
+		if (isSpecialized() || isNoSuggestionsText()) {
```
sspanak, line 202: Check if this breaks the Duolingo report field and the Sonim search field.

sspanak: They are fine.

PR #734, opened 2025-02-17
next language icon replaced with textual languages indication

barszczmm:
Globe icon shown on T9 button when there are two or more languages is replaced by "\<current language code\> > \<next language code\>" textual indication. This has two advantages over icon:
- currently active language is always visible (language in status bar is not always available to read)
- next language that would be active after button long press is clearly indicated

Screenshot showing changed UI:

tt9-languages


barszczmm:
@sspanak Do you think it would be better (performance wise) to keep next language in new variable and update that variable whenever current language changes and when list of selected languages changes? Then `getNextLanguage()` method would just return that variable instead of calculating next language. Should I add such change?

barszczmm:
There are two reasons I wanted to introduce this change:
- The _"you would start typing, realize it's in the wrong language and switch"_  case is confusing for me and my change already helped me (I use my own fork on my phone). This is personal preference obviously and I would probably get used to default UI, but since I'm developer and I can adjust the keyboard to my needs then why not to :) Also for me it's confusing that language indication is there and it later disappears (I'm talking about language in status bar).
- The always visible lang indication is more common UI pattern found in other soft keyboards (GBoard, TouchPal), so it would probably be easier for new users to use TT9.

Regarding Kika inspired language selection dialog: I would only show this dialog when there are 3 or more languages, because it requires more effort to switch (two clicks/touches and dialog mental processing). In case of two languages I would stick to automatic switch on long press.

Regarding problems with my solution:
- Now I understand the issue with the text. When I was testing this I realized that length of text I introduced is almost the same as ABC text, so I thought it will be fine, but I just tested this on very small sizes and ABC is resizing nicely, but my text doesn't and it looks bad. Yes, the icon solution eliminates this issue.
- Thanks for pointing out problem with color, I didn't notice it. I will fix this.

All in all I understand that you may not want this proposed UI change (especially when you have plans to add language selection dialog) and I'd understand if you reject this PR. I will just keep this change in my own fork for my usage then :) Anyway thanks for answering and for all the good work you do with this project. I'm very glad I found TT9 and I'm grateful for your work. I'm planning to extend my fork for my needs, so you may expect some more PRs in future, maybe you'll find some of my changes useful. Some features I will probably work on are:
- ability to use two languages at the same time (without the need of manually merging languages)
- ability to configure placement of each functional button (I use shift and T9 buttons very often, so I'd like to have them on right side, on the other hand I almost never use filter and editing buttons, so I could have them on left side - I think allowing users to reorder buttons may be useful)
- ability to disable resizing of the keyboard by dragging settings button or status bar - I resized the keyboard by mistake few times and I don't need this feature at all, so I'd like to be able to turn it off

sspanak:
> The "you would start typing, realize it's in the wrong language and switch" case is confusing for me and my change already helped me (I use my own fork on my phone). This is personal preference obviously and I would probably get used to default UI, but since I'm developer and I can adjust the keyboard to my needs then why not to :) Also for me it's confusing that language indication is there and it later disappears (I'm talking about language in status bar).

Now I remember that switching the language on-the-fly was actually requested by another user. So, another person finds this useful, but you find it confusing. We come back with the common "why don't you add feature X?" question. I simply can't make everyone happy. I like the way it works now, so I must reject this PR. Of course, you are welcome to use your mod. I would recommend minimizing the code changes, to make it easy to stay up-to-date with `master`.

> Regarding Kika inspired language selection dialog: I would only show this dialog when there are 3 or more languages, because it requires more effort to switch (two clicks/touches and dialog mental processing). In case of two languages I would stick to automatic switch on long press.

Exactly. Only display it when there are 3 or more languages. I haven't planned it though. I just came up with the idea recently.

> Anyway thanks for answering and for all the good work you do with this project. I'm very glad I found TT9 and I'm grateful for your work. I'm planning to extend my fork for my needs, so you may expect some more PRs in future, maybe you'll find some of my changes useful.

Enjoy! Btw, you are also welcome to make improvements to Polish language. It is one of the Slavic languages with a not-so-good dictionary.

> ability to disable resizing of the keyboard by dragging settings button or status bar - I resized the keyboard by mistake few times and I don't need this feature at all, so I'd like to be able to turn it off

Do this with a nicely looking code like in this PR, add the option on the Appearance screen, and I will merge it in the next version.

Btw, if you just slide down the keyboard, the effect is not permanent. It will go back to normal when you go to another app. Also, you can double tap the status bar to toggle between a small and a large layout. These are one of the most useful and cool features to me. I have a 4.5" phone and sometimes I just can't see anything because of the keyboard. It may not be so useful on a large screen though.

> ability to configure placement of each functional button (I use shift and T9 buttons very often, so I'd like to have them on right side, on the other hand I almost never use filter and editing buttons, so I could have them on left side - I think allowing users to reorder buttons may be useful)

I am not against this feature, but it requires too much work. And I do not have so much free time. Perhaps, it can be achieved by inflating a custom layout to avoid messing up the SoftKeys too much. If you do it, I am up for reviewing and potentially merging it.

> ability to use two languages at the same time (without the need of manually merging languages)

I have no idea how to do this. I mean, I can make TT9 search in two DB tables, but I don't know how to make it guess which language do you prefer at the moment. Besides this, you have to think about performance. Even now, a single search may take 100 ms on low-end devices. With the current architecture, the time will be almost double, which will cause significant typing lag. 

And even if you solve the technical problems, you will have to think about displaying the letters for multiple alphabets, as well as how to handle right-to-left and left-to-right languages at the same time. And there are probably more small problems. I don't even want to think about it and I have not considered such feature. Nevertheless, I would be curious to see a working solution.

PR #762, opened 2025-03-10
More robust voice input on Android 12+

sspanak:


app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -75,7 +75,14 @@ private void onVoiceInputStopped(String text) {
 
 
 	private void onVoiceInputError(VoiceInputError error) {
-		if (error.isIrrelevantToUser()) {
+		if (error.isLanguageMissing()) {
```
sspanak, line 78: On Android 12, just print the error message because automatic download is not possible

PR #765, opened 2025-03-19
Pinyin

sspanak:

PR #766, opened 2025-03-20
small fixes 49

sspanak:


app/src/main/java/io/github/sspanak/tt9/preferences/settings/SettingsTyping.java
```diff
@@ -30,6 +30,10 @@ public String getDoubleZeroChar() {
 		return  character.equals("\\n") ? "\n" : character;
 	}
 
+	public boolean getPredictiveMode() {
```
sspanak, line 33: Ensure this does not conflict with `hasABC` after merging Chinese

app/src/main/java/io/github/sspanak/tt9/ui/tray/SuggestionsBar.java
```diff
@@ -26,6 +26,7 @@
 import io.github.sspanak.tt9.util.chars.Characters;
```
sspanak, line 26: Also disable animation using `mView.setItemAnimator(null);` when there are more than 20 suggestions.

app/src/main/java/io/github/sspanak/tt9/preferences/settings/SettingsTyping.java
sspanak: Will take of this in #765

PR #792, opened 2025-04-25
System tray to display the current input mode.

MysteriousBlob:
This uses the current system tray icon to display the current input mode:
Robot if English/predictive, En if abc, Dial if 123.

It helps you know the current input mode if the screen layout is invisible or didn't show up for whatever reason.


app/build.gradle
```diff
@@ -116,7 +116,8 @@ android {
 			minifyEnabled true
 			shrinkResources true
 			proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
-		}
```
sspanak, line 119: Unnecessary code change.

app/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:tools="http://schemas.android.com/tools"
-	android:versionCode="1027"
```
sspanak, line 3: Unnecessary code change.

app/src/main/java/io/github/sspanak/tt9/languages/LanguageDefinition.java
```diff
@@ -169,7 +169,7 @@ private void setProperty(@NonNull String line) {
 	 */
 	private boolean setLayoutEntry(@NonNull String line) {
 		if (!inLayout) {
-			return inLayout = "layout:".equals(line);
+			return inLayout = "layout:\r".equals(line);
```
sspanak, line 172: As I mentioned in #793, this happens because you have not configured Git or Android Studio correctly to format all files correctly. It is not a bug with the Java code.

You have also pushed a lot `.zip` files that probably contain `\r` characters too. You must not commit such files, because they will break many other things.

app/src/main/java/io/github/sspanak/tt9/ime/UiHandler.java
```diff
@@ -56,7 +56,13 @@ public void initUi(InputMode inputMode) {
 
 
 	protected void setStatusIcon(InputMode mode) {
-		if (!InputModeKind.isPassthrough(mode) && settings.isStatusIconEnabled()) {
+		if (InputModeKind.isPredictive(mode) && settings.isStatusIconEnabled() && settings.isStatusIconTypeModeEnabled()) {
```
sspanak, line 59: The icons you have suggested do not indicate clearly the current state.

Second problem. How is "En" representative of ABC mode? TT9 currently supports 44 languages and more are coming. You should have added an icon for each language in its respective script.

When converting text to icons it often becomes disproportional on different devices. TT9 used to have icons for each language in the past, but I removed them all because of this problem. See the screenshots below to see what I mean.
Screenshot_20250427-143119
Screenshot_20250427-143055

sspanak (commented):
Thank you for the effort. Unfortunately, this PR has so many issues, that I must reject it.

MysteriousBlob:
Thank you very much for reviewing!
I am sorry for taking your time, this was a valuable experience for me.

PR #811, opened 2025-06-03
Final final 50

sspanak:


sspanak (commented):
Also, must check why the documents are duplicated in `docs/` and `help/` folders and leave only one of them.

PR #914, opened 2025-11-10
Add real-time transcription and proper capitalization to voice input

shmu1i:
## Summary
- Voice input now displays text as you speak instead of waiting until the end
- Provides immediate feedback with real-time transcription
- Proper capitalization for sentence beginnings and proper nouns

## Implementation
- Enabled partial results in speech recognition intent
- Implemented `onPartialResults()` callback in VoiceListener
- Used Android's composing text API to smoothly replace previous text without duplication
- Capitalization is handled automatically by the speech recognizer using `LANGUAGE_MODEL_FREE_FORM`

## Test plan
- [x] Test voice input with various phrases
- [x] Verify text appears in real-time as you speak
- [x] Verify no character duplication
- [x] Verify proper capitalization
- [x] Test final result is committed correctly

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
+++ b/app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
@@ -6,13 +6,11 @@ import android.view.KeyEvent;
 import io.github.sspanak.tt9.R;
 import io.github.sspanak.tt9.ime.modes.helpers.AutoTextCase;
 import io.github.sspanak.tt9.ime.modes.helpers.Sequences;
-import io.github.sspanak.tt9.ime.modes.InputMode;
 import io.github.sspanak.tt9.ime.voice.VoiceInputError;
 import io.github.sspanak.tt9.ime.voice.VoiceInputOps;
 import io.github.sspanak.tt9.ui.dialogs.RequestPermissionDialog;
 import io.github.sspanak.tt9.util.Logger;
 import io.github.sspanak.tt9.util.Ternary;
-import io.github.sspanak.tt9.util.Text;
 
 abstract class VoiceHandler extends TypingHandler {
 	private final static String LOG_TAG = VoiceHandler.class.getSimpleName();
@@ -104,33 +102,17 @@ abstract class VoiceHandler extends TypingHandler {
 	}
 
 
-	private String applyAutoCapitalization(String str) {
-		if (str == null || str.isEmpty() || autoTextCase == null) {
+	private String autoCapitalize(String str) {
+		if (autoTextCase == null) {
 			return str;
 		}
 
-		StringBuilder output = new StringBuilder(str.length());
-
-		for (String word : str.split(" ")) {
-			Text text = new Text(mLanguage, word);
-			int textCase = autoTextCase.determineNextWordTextCase(mLanguage, mInputMode.getTextCase(), inputType.determineTextCase(), textField, "", beforeSpeech + output);
-
-			String correctedWord = switch (textCase) {
-				case InputMode.CASE_UPPER -> text.toUpperCase();
-				case InputMode.CASE_CAPITALIZE -> text.capitalize();
-				case InputMode.CASE_LOWER -> text.toLowerCase();
-				default -> text.toString();
-			};
-
-			output.append(correctedWord).append(" ");
-		}
-
-		return output.toString();
+		return autoTextCase.adjustParagraphTextCase(mLanguage, str, beforeSpeech, mInputMode.getTextCase(), inputType.determineTextCase());
 	}
 
 
 	private void onVoiceInputStopped(String text) {
-		onText(applyAutoCapitalization(text), false);
+		onText(autoCapitalize(text), false);
 		resetStatus();
 		 if (!mainView.isCommandPaletteShown()) {
 			 mainView.render(); // re-enable the function keys
@@ -139,7 +121,7 @@ abstract class VoiceHandler extends TypingHandler {
 
 
 	private void onVoiceInputPartial(String text) {
-		textField.setComposingText(applyAutoCapitalization(text), 1);
+		textField.setComposingText(autoCapitalize(text), 1);
 	}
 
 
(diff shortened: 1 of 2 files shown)


shmu1i:
I tested this with multiple sentences and seems to be working with all sorts of combinations

app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -84,27 +87,100 @@ protected void stopVoiceInput() {
...
 
 
+	private void onVoiceInputPartial(String text) {
+		Logger.d(LOG_TAG, "Voice input partial: " + text);
+
+		// Skip empty partial results
+		if (text == null || text.isEmpty()) {
+			return;
+		}
+
```
sspanak:
```suggestion
```

1. Let's cut down on logging a bit. You are going to see the result on the screen anyway, so this shouldn't be necessary.
2. Both `textField.setComposingText()` and `applyAutoCapitalization()` know what to do with an empty or null input, so this extra validation here is not necessary.

app/src/main/java/io/github/sspanak/tt9/ime/voice/VoiceInputOps.java
```diff
@@ -200,6 +204,13 @@ private void onStop(ArrayList<String> results) {
 	}
 
 
+	private void onPartial(ArrayList<String> results) {
+		String text = results.isEmpty() ? null : results.get(0);
+		Logger.d(LOG_TAG, "Partial result: " + text);
+		onPartialResult.accept(text);
```
sspanak:
```suggestion
		if (!results.isEmpty()) {
			onPartialResult.accept(results.get(0));
		}
```

If there are no results, there is no point in calling `onPartialResult`, because... there are no results.

app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -84,27 +87,100 @@ protected void stopVoiceInput() {
...
+				if (TextTools.isStartOfSentence(before)) {
+					result.setCharAt(i, Character.toUpperCase(c));
+				}
+			}
+		}
+
+		return result.toString();
+	}
+
+
```
sspanak: There is already a class called `AutoTextCase` that is supposed to handle all such operations based on the settings. I don't like the fact that you are duplicating the logic here. It isn't maintainable. 

I would suggest getting the text "before", and creating an `AutoTextCase` in `onVoiceInputStarted()`,  then using them here. Something like:

```suggestion
	private String applyAutoCapitalization(String str) {
		if (str == null || str.isEmpty() || autoTextCase == null) {
			return str;
		}

		StringBuilder output = new StringBuilder(str.length());

		for (String word : str.split(" ")) {
			Text text = new Text(mLanguage, word);
			int textCase = autoTextCase.determineNextWordTextCase(mLanguage, mInputMode.getTextCase(), inputType.determineTextCase(), textField, "", beforeSpeech + output);

			String correctedWord = switch (textCase) {
				case InputMode.CASE_UPPER -> text.toUpperCase();
				case InputMode.CASE_CAPITALIZE -> text.capitalize();
				case InputMode.CASE_LOWER -> text.toLowerCase();
				default -> text.toString();
			};

			output.append(correctedWord).append(" ");
		}

		return output.toString();
	}
```

This will also require a small modification to `AutoTextCase.determineNextWordTextCase()`. It will have to optionally take the `beforeCursor` String, then use it on line 77:

```java
final String before = beforeCursor == null ? textField.getStringBeforeCursor() : beforeCursor;
```

app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -84,27 +87,100 @@ protected void stopVoiceInput() {
...
 	private void onVoiceInputStopped(String text) {
-		onText(text, false);
+		Logger.d(LOG_TAG, "Voice input stopped: " + text);
+		onText(applyAutoCapitalization(text), false);
+		shouldCapitalizeVoiceInput = false;
```
sspanak:
```suggestion
		onText(applyAutoCapitalization(text), false);
```

app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -84,27 +87,100 @@ protected void stopVoiceInput() {
...
 	private void onVoiceInputStarted() {
+		// Determine capitalization once at the start based on current context
+		shouldCapitalizeVoiceInput = false;
+		if (settings.getAutoTextCase()) {
+			String before = textField.getStringBeforeCursor();
+			shouldCapitalizeVoiceInput = before.isEmpty()
+				|| (settings.getAutoCapitalsAfterNewline() && before.endsWith("\n"))
+				|| TextTools.isStartOfSentence(before);
+		}
+
```
sspanak:
```suggestion
```
This may finish executing after the first spoken word has been set as a composing text. Instead, do the initialization in `toggleVoiceInput()` not here. Right before `voiceInputOps.listen(...)`, add the following lines:

```
		autoTextCase = new AutoTextCase(settings, new Sequences(), inputType);
		beforeSpeech = textField.getStringBeforeCursor();
```

where `autoTextCase` and `beforeSpeech` are private class properties.

app/src/main/java/io/github/sspanak/tt9/ime/VoiceHandler.java
```diff
@@ -9,10 +9,12 @@
 import io.github.sspanak.tt9.ui.dialogs.RequestPermissionDialog;
 import io.github.sspanak.tt9.util.Logger;
 import io.github.sspanak.tt9.util.Ternary;
+import io.github.sspanak.tt9.util.TextTools;
 
 abstract class VoiceHandler extends TypingHandler {
 	private final static String LOG_TAG = VoiceHandler.class.getSimpleName();
 	protected VoiceInputOps voiceInputOps;
+	private boolean shouldCapitalizeVoiceInput = false;
```
sspanak: Based on my other suggestions:

```suggestion
	private AutoTextCase autoTextCase;
	protected VoiceInputOps voiceInputOps;
	private String beforeSpeech = "";
```

shmu1i:
Ok implemented all your suggestions. Tested and working great

sspanak:
> Ok implemented all your suggestions. Tested and working great

Excellent! I've decided to move the logic from `autoCapitalize()` to the `AutoTextCase` class after all. There are no functional changes otherwise. I believe this is good to go now!

sspanak (approved):

PR #980, opened 2025-12-28
Languages 59 2

sspanak:


Pedrofsky:
I've added a few more words in the issue, I hope it's not too late.

PR #1034, opened 2026-01-29
prevented unnecessary text analysis when pressing backspace

sspanak:


sspanak:
The final result does not reduce the getString operations. They seem to be fairly optimal even now. Rejecting.

PR #1043, opened 2026-02-07
Add 4 new English words + change capitalization of two

ngoomie:
- Added 'cis': science term (also shows up in compounds as 'cis-' sometimes) + math term + used as an abbreviation for 'cisgender'
- Added '-ing': used when tacking 'ing' onto words that you wouldn't normally see with an 'ing' on the end (like turning a word that's usually only a noun into a verb) -- I do wonder if it might make more sense to just put 'ing' in instead, without the dash? That was almost what I did first
- Added 'convo': I dunno if this one is common enough to be included, I use it a lot in very casual speech + people I know do, but I could have a skewed view of how common it is
- Added 'filesystems': singular 'filesystem' is already here so I thought it was weird that the plural wasn't included
- Changed 'MASH' to 'mash': I think mash as a normal verb/noun is probably more common than MASH, the acronym. 'mash' as shorthand for mashed potatoes, or as mashed animal feed...
- Changed 'Colon' to 'colon': technically these are different words, capital-C being a surname, locaps-c being the body part or the punctuation. I think probably the latter usage is more common than the former?
- Removed 'zoot': this word is mostly pretty wildly uncommon nowadays so I was surprised to see it in the default dictionary at all. I *do* use the similar-but-unrelated word 'zooted' in very casual speech a lot (which is why I noticed it) but I certainly wouldn't think of it as a common enough word to include. See also: [this Google n-gram viewer query for 'zoot suit' and lone 'zoot'](https://books.google.com/ngrams/graph?content=zoot+suit%2Czoot&year_start=1800&year_end=2022&corpus=en&smoothing=3&case_insensitive=false) (this was a later addition hence it not being mentioned in the title, initially I wasn't sure enough to actually remove it, but I figure if you think it should stay then you can just not include that part of my merge req)

Sorry if it was weird to do these in multiple commits. I also don't know if I did the priorities right, I assumed lower number = shows up later in the list? Sorry too if I got sorting wrong.

### DIFF
--- a/app/languages/dictionaries/en-utf8.csv
+++ b/app/languages/dictionaries/en-utf8.csv
@@ -941,7 +941,7 @@ god	124
 hoe	72
 Goi
 IMG	79
-Ing
+ing
 hog	87
 Hol
 ink	109
@@ -1782,7 +1782,6 @@ wyd
 -per	10
 -ROM	10
 -toc	10
--ing
 AABB
 ACCA	70
 ACCC	70


beackers:
Also @ngoomie : 99% of my work is done in multiple commits, so whether that's bad form or not, I'm with you :)

sspanak:
Thanks taking time improving the dictionary!

I agree about the new words and the text case corrections. There are words wrongly in uppercase in all languages, but it is simply impossible for me to find and correct them. Please, do report such words in the future.

However, I disagree about removing "zoot". It has a frequency of 60, so it isn't so uncommon. Let's keep it.

Now, as for "-ing", could you provide examples of what you need to type? Adding "-ing" means you will be able to type words like this "someword-ing", while I suppose you are looking to type "somewording"?

Don't worry about the number of commits. As a minimum, I'll also generate the `.zip` file for downloading. The text file is useless on its own.

Ping @ngoomie.

ngoomie:
@sspanak The only specific example I can think of is I recently said "housewife-ing" when trying to shorthandedly refer to *the things housewives do*, since "housewifing" isn't really a word and also looks slightly off (though I later figured out "housewifery" is, which sort of means what I meant) — that's kind of generally how I tend to use it + how I see others use it too

ngoomie:
For the hell of it I went through a couple large or old Discord servers I'm in, and found:
- many instances of '[TV show/videogame/etc]-ing' (meaning "to watch/play [media in question]"), or similar like '[company name]-ing' (to refer to someway handing money to that company)
- [food]-ing (usually "making [food]", sometimes "eating [food]")
- DJ-ing (I'm sure "DJing" must be much more common though, but it came up a few times)
- Ctrl-F-ing
- ooh-ing and ahh-ing; aww-ing; hee-ing and haw-ing/hehe-ing and haha-ing; hum-ing and haw-ing; etc.
- carpal tunnel-ing ("getting bogged down by carpal tunnel") or other examples of multi-word nouns having -ing tacked on
- accidentally-ing ("doing something accidentally" when said mistake is obvious from context, f'ex I once said "accidentally-ing over my 2TB drive" to refer to accidentally overwriting said HDD)
- a few instances of '[emoji]-ing' or the same with text faces like XD (usually meant as someone using that emoji/text face a lot)
- I saw a couple instances of phrases or even individual words (picked from a larger phrase) that already ended with an 'ing' having an additional -ing tacked on. a notable one I saw a few times in a local Discord was "notwithstanding-ing" (referring to invoking of the "notwithstanding clause" from the Canadian Charter of Rights & Freedoms, due to that happening a few times in Alberta recently)

Also worth a mention: I guess it's somewhat common to use 'ing instead of -ing a lot depending on the context... like I saw @'ing a lot more than @-ing, or sometimes Ctrl-F'ing instead of with a dash, or actually *all* of the emoji examples were with an apostrophe, as in 😂'ing. So maybe it'd be good to just put in bare 'ing' on its own instead. I almost forgot too that "Ing" is also a surname, so then putting it in without the dash would make it easier to type that surname

ngoomie:
So I guess "Ing" the surname was already in there. Weird then that I never saw it when typing out 464 until I added "ing" to my personal dictionary :?

sspanak:
Yes, there may be other issues too. I'll double check before merging. Also, I would like to keep "zoot".

ngoomie:
I'm not so good with git and don't really know how to undo one commit while keeping the contents of later ones so I'll have to look that up later, or maybe just leave it to you?

beackers:
Probably most simple way is to just reinsert "zoot" as a new commit.

"zoot 60" appeared between "wont 50" and "Zora 27", so that's where I'd slot it back in.

For future reference, I think [this StackOverflow](https://stackoverflow.com/questions/15772134/can-i-delete-or-undo-a-git-commit-but-keep-the-changes) question should kinda answer what you're wondering. In your case, you'd probably want to reset to the "zoot" commit, then undo the change on zoot, then recommit... but tbh committing a reinsert would probably be faster.

sspanak:
I'll add it back, don't worry.

PR #1053, opened 2026-02-13
add latin

beackers:
I saw TT9 has Greek in it, so since I'm a Latin nerd, I was like "hmm, let me see". I use TT9 enough for typing Latin assignments as well, so... :)

I did get Codex by OpenAI to write this, and its notes are in `laWordlistReadme.txt`. Let me know if something needs to get changed.


sspanak:
Thanks for contributing!

Just to clarify, the Greek dictionary is not Ancient Greek. It is modern Greek used nowadays in Greece.

Either way, I am OK with adding dead or fictional languages, just for fun. But, if possible for these languages, I would like to add big dicitionaries that permit typing a lot of words in different contexts. For Latin, this should be easy - a lot of dictionaries exists today, and they can be easily incorporated in TT9. Unfortunately, I am quite busy now, so I can not spend time on this myself. On the other hand, you seem to have a lot of free time, and if you are willing to make some more effort, I will include Latin in the app. :slightly_smiling_face: 

I suggest that we use [this wordlist](https://www.winedt.org/dict.html). It contains more than 1 million words, so it should produce very nice typing experience. It may require some cleaning, e.g. remove the single letters, remove any words with corrupted or non-Latin letters, but it should mostly be fine.

As for the macrons, the word list from Winedt doesn't contain any. I would recommend actually installing and using the latin-macronizer tool you have found to make the dictionary nicer.

After that, I can build an APK for you to do some real-world testing, and if it feels alright, I'll merge it an publish it.

Go for it!

app/languages/definitions/Latin.yml
```diff
@@ -0,0 +1,15 @@
+locale: la
```
sspanak: TT9 does not support 2-letter locales, except for English.
```suggestion
locale: la-VA
```

app/build.gradle
```diff
@@ -48,6 +48,8 @@ tasks.register('buildDictionaryDownloads') {
 tasks.register('copyDownloadsToAssets', Copy) {
 	from DICTIONARIES_DOWNLOAD_DIR
 	include '**/*.zip'
+	// Keep Latin as a downloadable add-on language; do not bundle it in the full APK.
+	exclude 'la-utf8.zip'
```
sspanak: Please don't do this. The full version is intended to come with all languages. It does not have the Internet permission and it can not download anything.
```suggestion
```

docs/dictionaries/laWordlistReadme.txt
```diff
@@ -0,0 +1,18 @@
...
+https://github.com/Alatius/latin-macronizer
+
+Dictionary file used for conversion:
+https://raw.githubusercontent.com/Alatius/latin-macronizer/master/macrons.txt
+
+Upstream license:
+- GPL-3.0 (repository license file)
+
+Conversion notes for TT9:
+- Parsed the macronized inflected-form column (4th TAB-separated field), which includes full forms
```
sspanak: Please, don't parse the files. You may miss some edge cases and produce wrong results. Instead install and use the latin-macronizer as I mentioned in my first comment.

Alternatively, use the online version of the same tool with the appropriate settings, and convert the word list from Winedt in portions of 50000 chars (or less). Just be careful not to send too many requests and bring down their website. Give it a couple of seconds between each try.

app/languages/definitions/Latin.yml
```diff
@@ -0,0 +1,15 @@
...
+dictionaryFile: la-utf8.txt
+name: Lingua Latina
```
sspanak: Java should be able to generate a proper name. Let's specify it manually only if it is unclear, or if it fails.
```suggestion
```

app/languages/definitions/Latin.yml
```diff
@@ -0,0 +1,15 @@
...
+iconABC: ic_lang_latin
```
sspanak, line 3: `iconT9` is mandatory. Please add a new entry to `scripts/iconz/generate-all.sh`, generate the icons and put them in the `res` directory. 

`StatusIcon.java` also needs to be updated, but it will happen automatically, when you build the project with Android Studio, gradle, or whatever you are using. Just don't forget to include it in the pull request.

sspanak (changes requested):

beackers:
... wow I should have caught 90% of that. My apologies for submitting that.

I do indeed have quite a bit of free time, I'll get right to it 👍🏻

beackers:
@sspanak 
Okay, big fork in the road. I had a look at the Winedt Latin dictionary... and you're right, it doesn't contain macrons. Which, Chat is telling me a lot of Latin people don't type macrons, so I won't bother that.

The bigger question is that the dict is also huge. 29.517 MB, and 1243950 words. For its size, it does have the Latin I study and a lot of it that I haven't...

It doesn't appear to have duplicates, I haven't looked the thing over completely, but it does look like it's not unnecessarily duplicated (for instance: -que is a suffix basically meaning "and <that word"; que and \*que appear once in the dictionary).

Do you want me to include the whole 29 MB or try to trim it down to what most classical Latin folks will actually use?

sspanak:
> it doesn't contain macrons. Which, Chat is telling me a lot of Latin people don't type macrons, so I won't bother that.

From what I know, Romans were trying to find ways to mark long vowels. They felt the need to mark them, and I can see how adding macrons with modern-day fonts is useful, that's why I suggested using that tool to macronize the words. Also, from what I know, people who refuse to use macrons are hardcore fans who think they know everything. :smile:

> The bigger question is that the dict is also huge. 29.517 MB, and 1243950 words. For its size, it does have the Latin I study and a lot of it that I haven't...
> Do you want me to include the whole 29 MB or try to trim it down to what most classical Latin folks will actually use?

TT9 can handle even 2 million words, don't worry about it. See, in T9 keyboards, you must have as many words as possible, otherwise, you have to compose them letter by letter, which is painfully slow. And even if you don't know some word, maybe someone else knows it and may want to type. Also, you may want to quote or copy some text, even if you don't know all the words. You need to have them for nicer typing experience.

So, bring it on! The more, the better.

> It doesn't appear to have duplicates, I haven't looked the thing over completely, but it does look like it's not unnecessarily duplicated

Yes, these dictionaries are quite good. They rarely contain misspelled words, garbage words and whatnot.

> -que is a suffix basically meaning "and <that word"; que and *que appear once in the dictionary).

I am not quite sure what you mean by that. There are quite a few words that end with "-que". Maybe, I misunderstood you...

beackers:
I think I've got the ducks in a row now.

Unfortunately I'll have to leave out macrons, at least for now. Turns out that the macronizers out there are tuned for making historical documents more readable, not dictionaries (even if I put in ~25000 words at a time), and they do such a great job at it based on contextual clues. With a dictionary, the only context for the _n_th word are the _n-1_ words behind it... and accordingly, all the tools I tried (especially on edge cases

beackers:
Whoops misclick.

especially on edge cases like _malo_ and _ma\_lo\__, _with evil_ and _I prefer_) either macronized incorrectly or didn't at all. So for the sake of preserving accuracy (and not having to go through all 1.24m words by hand) I'll leave it out for now and keep my eyes out for a macronizer for dictionaries.

sspanak:
I see. I had a similar problem a couple of times already. The most notable case was Russian, where they have "е" and "ё", but most people use them inconsistently, instead of following the grammar rules. Naturally, this means most word lists found on the Internet contain misspelled words. So, I've found a tool that knew how to correct these two letters, and in the ambiguous cases, where two different words could be written with "е" or "ё" (just like "malo" and "mālō"), I kept both variants, because both _are_ valid words. In my case, it was easy, because the tool had the option of producing one of the variants or both, and it didn't require context, it just processed according to the configuration. But, as I understand, this is not the case with the Latin macronizers.

So, in summary, if you can't make the macronizer tool(s) produce all valid word variants, ignoring the context, then we can go without macrons, I guess. They would have been valuable though. Dropping them is a bit sad.

maverick74:
Sorry for jumping into this... I just stumbled upon it! 
Is it just Classical Latin? Or also Vulgar Latin? Maybe a mix?
I know almost nothing about Latin but it's the language I would like the most to learn! Specially the Ecclesiastical Latin form!

Is the Ecclesiastical form also "included" in this? 🙂

(edit: Ecclesiastical Latin does not use macrons or apex and use acute instead, from what I was able to learn)

sspanak:
The readme contains a link to [this website](https://www.cyllenius.net/labium/content/view/21/49/). Here is the relevant information:
> The database was developed by Stefano Minozzi starting from the words listed in Lessico Fondamentale Latino by E. Riganti (Bologna, 1989). The project aims to expand the words contained in the database to the whole dictionary of Classical Latin, up to the Middle ages.

My guess is, it started as a Classic Latin database, but then it was expanded with more words, potentially Vulgar Latin (the words from the Middle ages). But my knowledge is also minimal, so I might be wrong.

Maybe @beackers has a better idea?

maverick74:
I also found this, that maybe of interest about the variants of Latin:

https://www.newadvent.org/cathen/09019a.htm

The [Ecclesiastical Latin](https://archive.org/details/dictionaryofeccl0000stel) dictionary

And Vatican's [Lexicon Recentis Latinitatis](https://archive.org/details/pavanetto-cletus-lexicon-recentis-latinitatis) 

I'm not aware of any license for the above dictionaries... (in fact...I was not able to open the links, but that's probably something with my phone... it's archive.org it's surely available!)

sspanak:
> I'm not aware of any license for the above dictionaries... (in fact...I was not able to open the links, but that's probably something with my phone... it's archive.org it's surely available!)

I can open them, but the books pages seem to be scanned images. This isn't really useful.

maverick74:
Is this stuck by something I can help with? 🙂

sspanak (dismissed):

sspanak:
> Is this stuck by something I can help with? 🙂

Well, I am working on two new versions simultaneously but I've unexpectedly hit a GitHub usage limit, so I am waiting for it to reset. Using 5 different AIs made me super productive! :smile: 

Other than that, I did some final tweaks, and from my perspective, it is good to go.

@beackers, the icon files were empty so I generated and committed new ones. Did you forget to commit them?

Also, I decided to remove the macrons from the language definition. Since the dictionary file no longer contains any words with macrons, having them in the definition is not necessary. Or shall we keep them for completeness?

Finally, I rebased and squashed everything into a single commit, because updating with the latest master became too difficult.

maverick74:
> Using 5 different AIs made me super productive!

525 🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤

I'm joking!!! I'm joking!!! 🤣🤣🤣

> it is good to go.

Nice 😎

maverick74:
A small correction @sspanak :

In Portuguese, Latin language is shown as "Latina" however that is only used if in context we're saying "Latin language" - in Portuguese: "lingua Latina"!

If we're talking about the language name - which is the case in tt9 - the correct way should be "Latim" (instead of "latina", as it is)

sspanak:
Are you looking at the original language name on top or the localized one at the bottom? On my devices it correctly displays as "Latina" in Latin, and "Latim" in Portuguese. Double-checked it on the Motorola, as well...

image

maverick74:
> Are you looking at the original language name on top or the localized one at the bottom? On my devices it correctly displays as "Latina" in Latin, and "Latim" in Portuguese. Double-checked it on the Motorola, as well...

Oh! I see!!! You're keeping the language name in it's original name!

My mistake!

All good! 😀

Issue #873, opened 2025-09-21
[Feature request] Add "developer" mode/options/menu/keybinds


beackers:
(Honestly by now I need to just learn kotlin and java and write this myself🤣)
Here's what I imagined: 
- In the settings / options of the app, have a menu for developer options.
- Under that menu, have a toggle basically controlling whether the dev options have any effect.
- Also under that menu, have a list of key mappings that are listened for while the toggle is active, like holding STAR and1 opens a menu for selecting keypresses like CTRL+S or CTRL+X, STAR and 2 for ALT, STAR + 3 for ctrl alt, etc. Have the key mappings be editable, so someone who uses CTRL S a lot more than, say, CTRL V could edit the bindings list to have CTRL S at the top of the list. 

My thought would be that this would be most useful when using apps that take those keybindings, like a lot of web browsers (I have Via, specifically). Also (for we who don't have touchscreens but have it anyways) in Termux, there's no way to hit the Ctrl, Alt, or any of those other buttons that they put in to press those keys while you press a letter on the keyboard. 

Let me know what you think!

sspanak:
I was going to go this way, I even added a Tab key at one point, but then realized that Android apps do not treat it like computer apps, so I gave up the idea. Have you actually checked, with the hardware keyboard, that enough apps react properly to these key combinations? Let's start with validating the idea.

beackers:
Termux (the big, obvious one)
I think most web browsers will support keyboard shortcuts. The one I primarily use, Via, definitely does, but the reason I say most is because even the stripped-down browser that's apparently such a generic Chromium remake I still don't know its' actual name supports keyboard shortcuts.  
Literally every text editor I've used. There's a lot on that list, but I've had to uninstall most of them for unrelated reasons.  
File manager (downloaded from F-Droid)  
So, at least from my limited experience, apps that devs would actually use with a keyboard could definitely use it. It's a fair point that most of these apps have buttons that make keyboard shortcuts unnecessary, but for flip phones in particular these buttons are a lot of the time extremely difficult (Material Files) or impossible (Termux) to reach with a d-pad. Some apps in particular I've noticed it's hard to just get out of a large text input section (Material Files, MarkSimply, Simple Markdown) in general, even to just exit without editing anything.  
[...]

PR #1055, opened 2026-02-14
developer-level keybinds

beackers:
Finally did it, after almost half a year (#873)...

Adds "developer commands". (I can change that to something else if you want.)

* User opens Settings and can toggle developer commands on/off
* With commands on, the 7 key (in the command palette) becomes the dev commands shortcut
* When 7 is pressed, panel opens showing a list of modifiers mapped to number keys.
* User selects modifier(s) and presses POUND, then inputs a letter.
* App presses key + modifiers.

> STATUS: **hasn't been tested yet. I'm working on that.**

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/ime/CommandHandler.java
+++ b/app/src/main/java/io/github/sspanak/tt9/ime/CommandHandler.java
@@ -41,13 +41,14 @@ abstract public class CommandHandler extends TextEditingHandler {
 			resetStatus();
 		}
 
-		if (!shouldBeOff() && mainView.isDeveloperCommandsShown()) {
-			onDeveloperCommand(key);
+		if (!shouldBeOff() && awaitingDeveloperComboKey) {
+			sendDeveloperCombination(key, repeat);
 			return true;
 		}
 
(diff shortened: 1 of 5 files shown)


beackers:
I'm down a phone, so testing is going to take a bit longer than expected.

sspanak:
Thank you very much for the time an effort you have put in this! I appreciate the dedication to explore and modify a codebase you were not familiar with, written in a language that you didn't know. That's no small achievement.

That said, I realize I may not have been clear enough before, so let me clarify now: I do not plan to include CTRL, ALT, or similar "hacker-style" keys in TT9. They don't align with the project's scope or the needs of its general user base.

Even in the example you mentioned, Termux already provides its own CTRL, ALT, and SHIFT keys, so duplicating that functionality inside TT9 would not add much practical value.

With that in mind, I'll have to decline this pull request. But don't let this discourage you. Do keep working on your mod. The beauty of the open-source software is exactly this: you are free to fork, experiment, and shape the project to match your own vision and needs. Keep building and learning!

beackers:
I absolutely appreciate the encouragement 👍🏻 Thanks very much and I'll take it to heart!

PR #1093, opened 2026-04-07
Add initial support for Macedonian based on other South Slav languages

boriseto:
My only issue that I couldn't figure out is adding a few missing letters within the virtual keyboard itself. For #5 I'm missing "кл *љ* м" and for #6 I'm missing "н *њ* оп"


app/languages/definitions/Macedonian.yml
```diff
@@ -0,0 +1,15 @@
+locale: mk-MK
+dictionaryFile: mk-utf8.csv
+iconABC: ic_lang_cyrillic
+iconT9: ic_lang_mk
+layout:
+  - [SPECIAL] # 0
+  - [PUNCTUATION_DE] # 1
+  - [а, б, в, г] # 2
```
sspanak: The layout is currently very inconsistent with the rest of the Cyrillic-based languages, and with the unwritten rules of T9 keyboards in general. Usually, the letters that have the same shape are on the same key. For example, both "a" and "ä" are on the 2-key, and both "o" and "ö" are on the 5-key; "c", "č", "ć" are on the 2-key; and all that is true, regardless of the actual alphabet ordering, or the sounds they make.

Also, the standard letters found in all languages, that use the same alphabet, are found on the same keys, so you always know that: 2=ABC or 2=АБВГ; 3=DEF or 3=ДЕЖЗ, and then come any special letters. This way you don't have to switch mentally, if you need to use multiple languages.

Keeping consistency is also important for new users, and for users who are learning a new language, and want to try typing in it. It makes little sense to force them learn a different layout.

And last but not least, many phones do not have Cyrillic inscriptions, or they don't have letter inscriptions at all. In this case, it would be extremely confusing if the keyboard produced different letters per key only in one language.

With the above in mind, I propose updating the layout like this:

```yaml
- [а, б, в, г, ѓ] # 2
- [д, е, ж, з, ѕ] # 3
- [и, ј, к, ќ, л, љ] # 4
- [м, н, њ, о, п] # 5
- [р, с, т, у] # 6
- [ф, х, ц, ч, џ] # 7
- [ш] # 8
- [] # 9 is not used, like in Serbian
```

I am also OK with rearranging based on the phonetic values of the letters:

```yaml
- [а, б, в, г] # 2
- [д, ѓ, е, ж, з, ѕ] # 3
- [и, ј, к, л, љ] # 4
- [м, н, њ, о, п] # 5
- [р, с, т, ќ, у] # 6
- [ф, х, ц, ч, џ] # 7
- [ш] # 8
- [] # 9 is not used, like in Serbian
```

But I can not accept your variant.

sspanak (changes requested):
My only comment is about the letter layout.

Other than that, great work! Thanks for contributing!

app/languages/definitions/Macedonian.yml
boriseto: About the layout, I actually had the users in mind to be honest. All the devices with t9 keyboards that support the language, have the layout I proposed. Also the Xperia keyboard that was quite popular here from that side has the same layout.
While I understand your logic about it and it makes sense, I wouldn't be able to explain to elderly users that have the muscle memory of them being there (there are keyboards that don't have the letters written on).

But again, if it's not acceptable, I can change it to your proposal I guess.

sspanak: Fair point. I agree it should be comfortable for the experienced users. But also, keep in mind, the layouts on Motorola, Sony, Samsung, Nokia and the other leading companies differed a bit. So, it really depends on what phone you had in the past.

Now, since TT9 is a Nokia-style keyboard, I'd rather stick with the Nokia standard. 

I have an old Nokia 3310 that has 20 languages, including Macedonian, and that I use for reference when I am not sure how to do something. I just powered it on after many years (still working!), and saw the layout is like this:
```yaml
- [а, б, в, г] # 2
- [д, ѓ, е, ж, з, ѕ] # 3
- [и, ј, к, л, љ] # 4
- [м, н, њ, о] # 5
- [п, р, с] # 6
- [т, ќ, у, ф, х] # 7
- [ц, ч, џ, ш] # 8
- [] # 9 is still empty
```

Maybe we can go with this layout?

boriseto: Yeap, that one actually makes a lot sense to me. I can change it after I'm back from my trips (probably somewhere over the weekend).
Another question would be is it possible in the t9 to add "hold for a letter"?
Like, for us „л“-„љ“ and „н“-„њ“ are separate letters (as is „ѕ“), however „ѝ“ and „ѐ“ are special cases, so I was thinking if it's possible to add a case in which holding the number where и or е are to give the special signed one?

sspanak: > Yeap, that one actually makes a lot sense to me. I can change it after I'm back from my trips (probably somewhere over the weekend).

No rush. I am also quite busy with other stuff, and also have limited Internet access right now, so all development is on pause for a while. I can do it as well, when I sort out my stuff.

> Another question would be is it possible in the t9 to add "hold for a letter"?
Like, for us „л“-„љ“ and „н“-„њ“ are separate letters (as is „ѕ“), however „ѝ“ and „ѐ“ are special cases, so I was thinking if it's possible to add a case in which holding the number where и or е are to give the special signed one?

No, unfortunately, there are technical limitations to that. "Hold for a digit" is the only possible option right now. Instead, the standard approach for accented letters like "ѝ" and "ѐ" is to have them as separate letters. It is how it works in French, Portuguese, Italian and other languages, when diacritics are absolutely mandatory for proper spelling.

boriseto: I've added the ѝ and ѐ in the respective numbers as possible last letter (they are not usually used, but sometimes needed). In the t9 input though it should be fine, right?

sspanak: Excellent! I'll review everything again, maybe reorder the word list for faster loading, and it should be good to go.

By the way, could you also please provide a link to the source where you got the words from? I would like everything in the app to be transparent and to give credit to everyone else's work.

boriseto: For that one, actually it's a python script (very old side project I did quite a few years back) and actually it's a from a Wikipedia backup dump and went with the script several times to strip some values that were nonsense - coincidentally, just before I got the email that the build failed I was looking into it again to see if the file is mostly stripped out of the words and a quick device test showed it was fine with the old layout.

boriseto: I can't find the original project, but it was for the Maliit keyboard for UBports and it was on gitlab then moved to github or something, so the PR for the dict isn't there (might not have made it to the main project at all too), so I'm unsure if I can find it.
I can share my python script if you want me to.

sspanak: I guess that makes you the author of the word list. And I understand I have your written permission to use it, and modify it in the future (if needed), and the permission extends to any potential enthusiasts who may want to update it (if they ever decide to).

boriseto:
Hey @sspanak , now that I'm temp back from travels, do you need anything else from me here?

sspanak:
Thanks for coming back to the issue! It's all good. I'll use the last layout we agreed on, I'll double-check if we missed something, and it should be good to go.

But before that, I would like to do a couple more features, and then I'll continue with adding new languages. It's just easier to organize the work this way.

boriseto:
When it comes to this one, no, however I might open a suggestion in the issues and we can have the discussion there.

sspanak:
> When it comes to this one, no, however I might open a suggestion in the issues and we can have the discussion there.

I mean, I am going to use your layout and your code as it is right now.

borismihailov-pfmintelligence:
Yes, this is fine the way it is right now.

I've read your documentation and FAQ about the reasoning about the layout and everything (after the initial discussion we had) and I was hoping to introduce (and maybe help with the development in the future when I have time) a custom layout setup wizard or something and didn't want to bring it to discussion here since it's been discussed in many topics, so it's not connected to this layout only.

sspanak:
> Yes, this is fine the way it is right now.
> 
> I've read your documentation and FAQ about the reasoning about the layout and everything (after the initial discussion we had) and I was hoping to introduce (and maybe help with the development in the future when I have time) a custom layout setup wizard or something and didn't want to bring it to discussion here since it's been discussed in many topics, so it's not connected to this layout only.

Feel free to open a new issue, and to experiment with the code. Have in mind it is not a trivial task though. The app is not designed to work this way, so it may require significant refactoring.

Also, you have posted from your work account. Just saying, if it matters which one you use to post here. :wink:

boriseto:
Yeah, noticed a bit later, but it isn't an issue, it was my decision to keep them separate for my own reasons. :)
Thanks though.

sspanak (approved):

Issue #659, opened 2024-10-27
Allow voice input with Futo Voice


vinceingithub:
Please consider adding an integration with Futo voice input as it is more privacy focused. I assume this app uses a Google API which isn't ideal.

sspanak:
TT9 uses the SpeechRecognizer API, so it can theoretically work with any voice input app, not necessarily only Google. For example, on Sonim phones, it works with their own voice assistant. Unfortunately, Futo Voice does not support that API (according to their readme), hence it can not integrate with TT9.

One more potential issue that I noticed is, the Whisper API, on which Futo is based, has poor support for many of the languages TT9 has. This means many TT9 users will not be able to benefit from having this alternative. 

All in all, I am not really convinced that integrating with Futo is worth the time and the effort at this time. Sorry to disappoint you!

rudism:
This would be useful for me as well.

There are other on-device voice inputs such as [Sayboard](https://github.com/ElishaAz/Sayboard) as well. I don't think it should require direct integration with specific keyboards--I'm not sure the details (they may not support the SpeechRecognizer API) but you configure them system-wide under Default Apps->Digital Assistant App->Voice Input which is what configures them to launch when clicking the microphone button from other applications, for example both Firefox and [Heliboard](https://github.com/Helium314/HeliBoard) detect it somehow and show a microphone button that launches the voice input app.

It may be a separate issue (or related to the Futo issue) but right now with Sayboard as my voice input assistant I get the error "Incompatible voice assistant" when I click the TT9 microphone button.

sspanak:
Firefox and Heliboard do not really use the voice inputs. Instead, when you press the "microphone" button, they activate the currently selected one and let it do the job itself.

Actually, thinking about it, it shouldn't be so difficult to do the same in TT9. The experience will not be seamless like now, but it will work for absolutely any voice input that you have installed, be it FUTO, Sayboard, and whatnot.

Here are the steps to do it, roughly:
1. Update the code that checks whether to show a microphone icon/button. In addition to checking for Google, also check if any other voice input method is available, with: `context.getPackageManager().queryIntentActivities(new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH), PackageManager.MATCH_DEFAULT_ONLY)`.
2. When the microphone is pressed in TT9, double check if another voice input has been installed in the background. Use `queryIntentActivities()` again.
    1. If not, use Google, like now.
[...]

PR #1165, opened 2026-07-14
Replace Google STT with offline Whisper voice input engine

JTDITech:
## Summary

This PR replaces Google Speech-to-Text with an offline Whisper-based voice input system using FUTO Whisper GGML.

## What changed

- Added a C++ + JNI bridge for the FUTO Whisper GGML engine
- Switched voice input to downloadable offline models
  - English Tiny
  - Multilingual Tiny (57 languages)
- Added automatic model selection based on the active language
- Blocked unsupported languages:
  - Irish
  - Toki Pona
  - Tamazight
- Updated voice hotkey behavior to force-show the keyboard window in passthrough fields
- Removed SSL certificate pinning
- Removed the bundled model asset

## Why

This improves voice input by making it offline-capable, reducing dependency on Google services, and enabling broader language support with language-aware model selection.

## Notes

- Users will need to download the required models
- Some languages are intentionally blocked due to unsupported model coverage

## Testing

- Verify voice input works with both English and multilingual models
- Confirm model auto-selection follows the active language
- Confirm unsupported languages are blocked correctly
- Test voice hotkey behavior in passthrough fields


sspanak:
Thank you for your effort. It must have been a significant effort to integrate a C++ library into an Android app. However, I must reject this PR for the following reasons:

1. I would like this project to remain as simple as possible, with minimum dependencies. Sort of, I don't want one more thing to break.
2. The biggest part of TT9 users are on Google Play. Removing the Google Voice Input would be a downgrade for them.
3. How about the users who do want to use Google Voice, even if they have a custom ROM? It is not fair to take it away from them.
4. This whole thing is actually not necessary. #659 (completed in v60.0) already adds support for any external voice input method. If you want to use Whisper, just install FUTO Voice, enable it from Android settings, and the next time you attempt to use voice input in TT9, it should show a popup dialog to select the desired voice input method. You can also install and use any other voice input app, you are not limited to just one.

Issue #1182, opened 2026-08-03
Custom emoji list


nclm:
Currently, TT9 allows you to input emoji from a “brief list” with the 1-key.

This is great, but is those are not the emojis you usually use, you are stuck and cannot insert more.

It would be great to provide an option to define the emojis including in the brief list(s), exactly like you can choose the Key Characters.

sspanak:
You can add emojis using the Key Characters setting, if you want to. Other than that, I do not plan to add any other options regarding emoji. Please, see #1183 for more info.

nclm:
Could there be a compromise, without the ability to modify the default ones, but with one extra emoji field, empty by default, that can be filled, up to X emojis? So that it doesn’t compete with your paid feature and keep it attractive, but still is practical for users without Google accounts and/or with ungoogled devices that ships with TT9 like the Dumber Mini. I have three languages on the phone, and I wouldn’t want different emojis per language.

Issue #1183, opened 2026-08-03
Full emoji list


nclm:
Currently, TT9 allows you to choose between two options for imputing emoji with the 1-key:
- No emojis
- Brief list (see #1182 about customising this)

There could be a third option:
- Full list

This would display a proper emoji picker, with good keyboard navigation, focused on a field that enables searching for emojis by name.

Alternatively, maybe it could partner with a third-party opensource emoji keyboard to switch to it instead if installed.

sspanak:
If you need to use the full emoji list, please install the app from Google Play, then activate Premium. Once you do this, you will get access to the third option in question. There is a proper emoji picker that can be used both with the touchscreen, and with the D-PAD (if your phone has one). You can switch between the emoji categories using the 0-9 keys (0 is for recently used).

The full list is not going to be available in the GitHub version.

nclm:
Thanks for your answer. I do not have a Google account and I use a Dumber Mini device that is ungoogled and ships with TT9 as a default. I understand that you are keeping this as a paid feature though.

PR #1184, opened 2026-08-03
Update Emoji lists

nclm:
I opened #1182 about being able to customise the emoji lists from within the app, but in the meantime, I thought why not proposing some of my most used emojis to the default lists.

Notably 😊, 😁, 😆, 😅, 🤞, 🤩, 🫂.

Then I got carried away and tried to make sense of it all:
- Each list is now exactly 10 emojis
- Changed some of the order to make it a little more logical and/or having more likely to be used emojis to the left
- Split the `emotions` list into `love` and `misc`, bringing it to 5 lists (50 emojis total)
- I filled the love list with 🫶, ❤️‍🔥, 💗, 💋 from https://emojipedia.org/most-popular
- Also added from there 🔥 to misc, as well as 🥵 to counter 🥶
- Only emoji I removed to fit neatly into 10 is 🍺, fun but maybe not universally used?

This is just a quick job and could be improved. Maybe some actual global stats for most used emoji could be used to choose the 50 or so most relevant? Maybe they could be ordered the same way they are in actual emoji pickers? But for now, this does the job for me. Feel free to approve it or not! 😊

### DIFF
--- a/app/src/main/java/io/github/sspanak/tt9/util/chars/Emoji.java
+++ b/app/src/main/java/io/github/sspanak/tt9/util/chars/Emoji.java
@@ -13,19 +13,23 @@ class Emoji extends Punctuation {
 	final private static ArrayList<ArrayList<String>> Emoji = new ArrayList<>(Arrays.asList(
 		// positive
 		new ArrayList<>(Arrays.asList(
-			"🙂", "😀", "🤣", "🤓", "😎", "😛", "😉"
+		  "🙂", "😊", "😁", "😆", "😅", "🤣", "😛", "😉", "🤓", "😎"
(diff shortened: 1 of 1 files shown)


sspanak:
I am not going to add any extra emojis to the GitHub version. Please, see #1183 for more info.

nclm:
I won’t be able to use the full emoji list as I don’t have a Google account, so I won’t be able to use the emoji I need on my device :(

I think having a slightly more complete and relevant default emoji list do not compete with your paid feature? It would still be very far from a full emoji picker and still makes your paid offer attractive. Maybe have a limit of 50 default emoji or something.

PR #1186, opened 2026-08-05
Add frequency for latin

coolcoder613eb:


sspanak:
@coolcoder613eb, that's fantastic thank you! Could you please also provide a link to the source where you got the word frequencies from?

coolcoder613eb:
I got them from here: https://github.com/lascivaroma/latin-lemmatized-texts

coolcoder613eb:
As well, how would you think enclitic suffixes should be handled? It's similar to english contractions, but doesn't have the punctuation marker. Things like -ne and -que (estne, visne, gladiusque, fluviusque).

sspanak:
> As well, how would you think enclitic suffixes should be handled? It's similar to english contractions, but doesn't have the punctuation marker. Things like -ne and -que (estne, visne, gladiusque, fluviusque).

The app doesn't understand grammar, so it doesn't know what suffixes are. The English contractions are treated differently exactly because they contain punctuation. But in the case of Latin, all of these must be included in the dictionary as separate words. There is no other way.

coolcoder613eb:
That would increase the size of the dictionary by quite a bit unfortunately, given for full coverage one would technically need a copy of every single noun form with -que, and every verb form with -ne... Is it possible to add custom handling for latin allowing -que, -ne, and -ve on all words?

sspanak:
It is true the dictionary size will grow significantly. But we don't have to do this. Since "que", "ne" and "ve" are already in the dictionary, you can just append them to any word without any special handling, for example:

1. Type "vis"
2. If you have "automatic space" enabled, press "clear filter" (D-PAD Down, swipe down OK, or press and hold the "Filter" key, depending on the device and layout). This will accept the word without adding an extra space after it. If you don't use "automatic space", just accept the word with D-PAD Center, with OK, or by tapping it (depending on your device).
3. Type the desired suffix.
4. Type a space, to proceed with the next word.

It may feel a bit odd at the beginning but it allows for ultimate flexibility. I think that's a great alternative to including all words.

PR #1193, opened 2026-08-11
Fix frequency for suffixes

coolcoder613eb:

PR #1194, opened 2026-08-11
Fix frequency for suffixes

sspanak:
Closes #1193


sspanak:
The original PR did not include the .zip download. Since I can not push from my computer to @coolcoder613eb's repo, I created a new branch to add the .zip file.

